PR #72, opened 2018-06-06
add sync task

bjfletcher:
It worked. Phew! I thought I'd make it easier by turning it into a `make sync`. Feel free to drop this if you don't like the look of it. :)

### DIFF
--- a/health/db-sync-state.js
+++ b/health/db-sync-state.js
@@ -37,20 +37,7 @@ module.exports = exports = new (class DBSyncStateCheck extends nHealthCheck {
 })({
 	businessImpact: 'The Syndication database\'s computed tables are not in-sync with the base table data and, as such, may be showing incorrect data to users.',
 	name: 'Syndication database data integrity',
-/*eslint-disable*/
-	panicGuide: `To force a reload of all the computed tables, run:
-
-\`\`\`
-	~$ psql --username \${PRODUCTION_DATABASE_USER_NAME}
-			--password \${PRODUCTION_DATABASE_PASSWORD}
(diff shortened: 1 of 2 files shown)


bjfletcher:
image

health/db-sync-state.js
```diff
@@ -37,20 +37,7 @@ module.exports = exports = new (class DBSyncStateCheck extends nHealthCheck {
...
-			--password \${PRODUCTION_DATABASE_PASSWORD}
-			--host \${PRODUCTION_DATABASE_HOST}
-			--port \${PRODUCTION_DATABASE_PORT}
-			--dbname \${PRODUCTION_DATABASE_NAME}
-			--command 'select syndication.reload_all();'
-\`\`\`
-
-Substituting the above \`\${PRODUCTION_DATABASE_*}\` placeholders with the correct values from vault.`,
-/*eslint-enable*/
+	panicGuide: 'you can sync them with the repo as follows: 1. ngage get-config --env prod 2. make sync',
```
leggsimon: This should explain that you should clone and cd into the repo first I think

bjfletcher: @leggsimon good point, let's not merge this PR... I think maybe it'll be nice to have a check early in `n-gage` to check whether .env is production, and prompt the user "is this what you want to load?"

leggsimon:
Thanks for this @bjfletcher but I'm going to close this in favour of #75 and would prefer to keep the `psql` command over `make` as it doesn’t require someone to clone the repo to run. Feel free to re-open if you have concerns.

PR #77, opened 2018-07-19
GDPR erasure screen

dapenguin:
Adds a very basic page, accessed via `/syndication/erasure`, where you can search records for the occurrence of email addresses and uuids in the Syndication DB. The page uses Single Sign-On authentication to allow access.

**Search Screen:**
image

**Results Screen:**
image

**Please Note:** This screen only searches for records. It does not delete any that are found.


server/controllers/erasure-search.js
```diff
@@ -0,0 +1,97 @@
+const { default: log } = require('@financial-times/n-logger');
+const operation = 'erasure-search';
+
+async function searchUserTableByUuid (db, uuid, requestedBy) {
+	try {
+		const userTableResults = await db.run(`SELECT * FROM syndication.users WHERE user_id = '${uuid}';`);
```
kavanagh: It looks like this could allow SQL injection. I don't know the Postgres `massive` library so it's hard to tell; looking at the docs it says the `run` method is deprecated and doesn't provide much info about guarding against injection.

server/controllers/erasure-search.js
```diff
@@ -0,0 +1,97 @@
...
+		log.info({ operation, subOp: 'search-user-table-by-uuid', requestedBy, uuid, userTableResults});
+		return { uuid, table: 'users', recordsFound: userTableResults.length };
+	} catch (error) {
+		throw error;
+	}
+}
+
+async function searchUserTableByEmail (db, email, requestedBy) {
+	try {
+		const userTableResults = await db.run(`SELECT * FROM syndication.users WHERE email = '${email}';`);
```
kavanagh: Ditto SQL injection.

server/controllers/erasure-search.js
```diff
@@ -0,0 +1,97 @@
...
+		log.info({ operation, subOp: 'search-user-table-by-email', requestedBy, email, userTableResults});
+		return { email, table: 'users', recordsFound: userTableResults.length };
+	} catch (error) {
+		throw error;
+	}
+}
+
+async function searchMigratedUsersTableByUuid (db, uuid, requestedBy) {
+	try {
+		const migratedUserTableResults = await db.run(`SELECT * FROM syndication.migrated_users WHERE user_id = '${uuid}';`);
```
kavanagh: Ditto

server/controllers/erasure-search.js
```diff
@@ -0,0 +1,97 @@
...
+}
+
+function getInvalidEmails (emails) {
+	return emails.filter(email => !/[A-Z0-9a-z._%+-]+@[A-Za-z0-9.-]+\.[A-Za-z]{2,64}/.test(email));
+}
+
+module.exports = async (req, res, next) => {
+	const requestedBy = req.cookies.s3o_username;
+	const db = res.locals.$DB;
+	const { erasureEmails, erasureUuids } = req.body;
```
kavanagh: Perhaps validate/sanitise the inputs here and throw when there are bad inputs (and return a 400 error to the client).

server/controllers/erasure-search.js
```diff
@@ -0,0 +1,97 @@
...
+function getInvalidUuids (uuids) {
+	return uuids.filter(uuid => !/[a-f\d]{8}-[a-f\d]{4}-[a-f\d]{4}-[a-f\d]{4}-[a-f\d]{12}/.test(uuid));
+}
+
+function getInvalidEmails (emails) {
+	return emails.filter(email => !/[A-Z0-9a-z._%+-]+@[A-Za-z0-9.-]+\.[A-Za-z]{2,64}/.test(email));
+}
+
+module.exports = async (req, res, next) => {
+	const requestedBy = req.cookies.s3o_username;
```
kavanagh: These responses shouldn't be cacheable by the client or proxies so we need to set `Cache-Control` here.

alicebartlett:
@dapenguin Hello, could you take a look at these comments or close this PR?

dapenguin:
@alicebartlett I'll try to address @kavanagh comments this week and get this merged as it would be useful to have.

server/controllers/erasure-search.js
dapenguin: This is currently using v3.1.0 of `massive`, and the example provided in [the docs for that version](https://gitlab.com/dmfay/massive-js/tree/v3.1.0#raw-sql) shows how to provide parameters to your query. Parameterised Queries are how Postgres protects against SQL injection.

I'll update the queries here to use parameters instead of string literals.

leggsimon: Just an FYI #163 upgrades massive

server/controllers/erasure.js
```diff
@@ -0,0 +1,3 @@
+module.exports = (req, res) => {
+	res.render('erasure');
```
leggsimon: this is so tiny, can it just be inlined?

server/controllers/erasure-search.js
```diff
@@ -0,0 +1,98 @@
+const { default: log } = require('@financial-times/n-logger');
+const operation = 'erasure-search';
+
+async function searchUserTableByUuid (db, uuid, requestedBy) {
+	try {
+		const userTableResults = await db.run('SELECT * FROM syndication.users WHERE user_id = $1;', [uuid]);
+		log.info({ operation, subOp: 'search-user-table-by-uuid', requestedBy, uuid, userTableResults});
+		return { uuid, table: 'users', recordsFound: userTableResults.length };
+	} catch (error) {
+		throw error;
```
leggsimon: should we not be logging these errors anywhere?

server/controllers/erasure-search.js
```diff
@@ -0,0 +1,98 @@
...
+	}
+}
+
+async function searchUserTableByEmail (db, email, requestedBy) {
+	try {
+		const userTableResults = await db.run('SELECT * FROM syndication.users WHERE email = $1;', [email]);
+		log.info({ operation, subOp: 'search-user-table-by-email', requestedBy, email, userTableResults});
+		return { email, table: 'users', recordsFound: userTableResults.length };
+	} catch (error) {
+		throw error;
```
leggsimon: ditto

server/controllers/erasure-search.js
```diff
@@ -0,0 +1,98 @@
...
+	}
+}
+
+async function searchMigratedUsersTableByUuid (db, uuid, requestedBy) {
+	try {
+		const migratedUserTableResults = await db.run('SELECT * FROM syndication.migrated_users WHERE user_id = $1;', [uuid]);
+		log.info({ operation, subOp: 'search-migrated-user-table-by-uuid', requestedBy, uuid, migratedUserTableResults});
+		return { uuid, table: 'migrated_users', recordsFound: migratedUserTableResults.length };
+	} catch (error) {
+		throw error;
```
leggsimon: ditto

server/controllers/erasure-search.js
```diff
@@ -0,0 +1,98 @@
+const { default: log } = require('@financial-times/n-logger');
+const operation = 'erasure-search';
+
+async function searchUserTableByUuid (db, uuid, requestedBy) {
+	try {
+		const userTableResults = await db.run('SELECT * FROM syndication.users WHERE user_id = $1;', [uuid]);
```
leggsimon: could this batch instead, so the query becomes
```sql
SELECT * FROM syndication.users WHERE user_id in ['…', '…'];
```

server/controllers/erasure-search.js
```diff
@@ -0,0 +1,98 @@
...
+		log.info({ operation, subOp: 'search-user-table-by-uuid', requestedBy, uuid, userTableResults});
+		return { uuid, table: 'users', recordsFound: userTableResults.length };
+	} catch (error) {
+		throw error;
+	}
+}
+
+async function searchUserTableByEmail (db, email, requestedBy) {
+	try {
+		const userTableResults = await db.run('SELECT * FROM syndication.users WHERE email = $1;', [email]);
```
leggsimon: could this batch instead aswell?

```sql
SELECT * FROM syndication.users WHERE email in ['…', '…'];
```

server/controllers/erasure-search.js
dapenguin: Apparently `IN` doesn't work with parameterised queries, but `ANY` does.

Only thing about doing that is that I would have to change how I output the results, and to be honest I don't think it's worth it for the small number of requests we'd be handling.

PR #84, opened 2018-09-06
Add Variant of resolve api which uses API KEY authentication

maxcbc:
- adds variant of resolve api which uses API KEY based authentication and a path param for contract id
 🐿 v2.10.2


Makefile
```diff
@@ -32,7 +32,7 @@ provision:
 	nht float --master --no-destroy --skip-gtg --testapp ${TEST_APP} --vault
 
 run:
-	nht run --local --https
+	nht run --https
```
leggsimon: I think you should be using `make run-local` instead so I'm not sure this should be removed.

server/middleware/get-contract-by-id-from-param.js
```diff
@@ -0,0 +1,17 @@
+'use strict';
+
+const getContractByID = require('../lib/get-contract-by-id');
+
+
+module.exports = (req, res, next) => {
+	const {contract_id} = req.params || {};
+	if (!contract_id) {
+		return res.sendStatus(400);
```
leggsimon: Could this maybe send a message as well like, `contract_id not provided` or something like that?

server/middleware/get-contract-by-id-from-param.js
```diff
@@ -0,0 +1,17 @@
...
+	}
+	getContractByID(contract_id)
```
leggsimon: does this need to be returned or awaited?

server/app.js
```diff
@@ -113,8 +114,9 @@ if (process.env.NODE_ENV !== 'production') {
 		cache,
 		apiKey
 	];
-
+	app.post('/syndication/contracts/:contract_id/resolve', [...middleware, getContractByIdFromParam], require('./controllers/resolve'));
```
leggsimon: I don't think you'd necessarily need to use the array spread here. I think you could add the getContractByIdFromParam as the 3rd param but 🤷‍♂️ don't care really

```js
app.post(
	'/syndication/contracts/:contract_id/resolve',
	middleware,
	getContractByIdFromParam,
	require('./controllers/resolve')
);
```

Makefile
maxcbc: why?

server/app.js
maxcbc: will ditch the spread

server/middleware/get-contract-by-id-from-param.js
maxcbc: Nope, not in a synchronous function.

server/app.js
maxcbc: it is so ditched

server/middleware/get-contract-by-id-from-param.js
maxcbc: done and dusted

Makefile
leggsimon: IDK

leggsimon: Cargo culting

server/middleware/get-contract-by-id-from-param.js
leggsimon: But `getContractByID` isn’t synchronous is it?

PR #91, opened 2018-10-19
Dont throw, log

maxcbc:
Rather than throw an error when an addendum doesn't have a matching asset, just log it

 🐿 v2.10.2

### DIFF
--- a/server/lib/reformat-salesforce-contract.js
+++ b/server/lib/reformat-salesforce-contract.js
@@ -1,6 +1,7 @@
 'use strict';
 
-const { ASSET_TYPE_TO_CONTENT_TYPE } = require('config');
+const {default: logger} = require('@financial-times/n-logger');
+const {ASSET_TYPE_TO_CONTENT_TYPE} = require('config');
 
 module.exports = exports = SFContract => {
 	const contract = {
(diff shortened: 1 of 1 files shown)


server/lib/reformat-salesforce-contract.js
```diff
@@ -16,46 +17,39 @@ module.exports = exports = SFContract => {
...
-				asset.content_type === addendum.content_type);
+	SFContract.assets
+		.filter(({assetType}) => assetType === 'Addendum')
+		.map(formatAsset)
+		.forEach(addendum => {
+			const asset = findMatchingAsset(addendum, contract);
 
 			if (!asset) {
-				throw new ReferenceError(`Asset not found for Addendum: ${JSON.stringify(addendum)}`);
+				logger.error(new ReferenceError(`Asset not found for Addendum: ${JSON.stringify(addendum)}`));
```
GlynnPhillips: Looks good. Thought that @leggsimon might have some opinions on, would it be good to set up a splunk alert based on this?

leggsimon: Yes it would, or logging to slack or something

PR #105, opened 2018-12-12
Fix user profile service api call

leggsimon:
So, the `/users/{user_id}/profile` API endpoint suddenly was deprecated, this meant that requests were being redirected to the new [`/users/{user_id}/profile/full`](https://developer.ft.com/portal/docs-membership-platform-api-user-api-get-users-userid-profile-full) endpoint. Unfortunately the scope of the access token we were sending was `profile_min` but the `/users/{user_id}/profile/full` endpoint requires `profile_max`.

This PR changes the user profile request to go to [`/users/{user_id}/profile/basic`](https://developer.ft.com/portal/docs-membership-platform-api-user-api-get-users-userid-profile-basic) which will accept the `profile_min` access scope.


robsquires:
🎯

PR #129, opened 2019-04-02
Switch to using n-mask-logger to mask PII

remybach:
🐿 v2.12.3

### DIFF
--- a/server/lib/logger.js
+++ b/server/lib/logger.js
@@ -1,0 +1,10 @@
+const MaskLogger = require('@financial-times/n-mask-logger').default;
+
+module.exports = new MaskLogger([
+	'email',
+	'password',
+	'contract_id',
+	'first_name',
+	'surname',
+	'x-api-key'
+]);
--- a/worker/sync/db-persist/spoor-publish.js
+++ b/worker/sync/db-persist/spoor-publish.js
@@ -2,7 +2,7 @@
 
 const path = require('path');
 
-const { default: log } = require('@financial-times/n-logger');
+const log = require('../../../server/lib/logger');
 const fetch = require('n-eager-fetch');
 
 const {
--- a/db/pg/index.js
+++ b/db/pg/index.js
@@ -2,7 +2,7 @@
 
 const path = require('path');
 
-const { default: log } = require('@financial-times/n-logger');
+const log = require('../../server/lib/logger');
 
 const massive = require('massive');
 const pgConn = require('pg-connection-string');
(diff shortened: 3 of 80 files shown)


server/lib/logger.js
```diff
@@ -0,0 +1,11 @@
+const MaskLogger = require('@financial-times/n-mask-logger').default;
+
+module.exports = new MaskLogger([
+	'email',
+	'password',
+	'contract_id',
+	'user_id',
```
richardLuko: why do you want to mask user id? I think it could be pretty useful for debugging

worker/sync/db-persist/spoor-publish.js
```diff
@@ -2,7 +2,7 @@
 
 const path = require('path');
 
-const { default: log } = require('@financial-times/n-logger');
+const log = require('../../../server/lib/logger');
```
richardLuko: wouldnt it be better to change the affected place, rather than updating all loggers?
do we apply same practice on all our services?

umbobabo:
@remybach that is fine for the masking purpose. 

My concern is how we can have the right information to do a proper (and possibly simpler) investigation in case of an issue, considering we won't be able to identify any user.
This is why I was suggesting a partial obfuscation of some field.
I think the application doesn't always log useful information for debugging like probably in this case for example https://github.com/Financial-Times/next-syndication-api/blob/17e7040994749ae86d9f8c5a8f461a572bade7cf/server/controllers/redshift.js#L21

That can be probably addressed in another PR though.
What do you think @liamkeaton @richardLuko ?

worker/sync/db-persist/spoor-publish.js
remybach: Not sure I understand what you mean, sorry...

liamkeaton:
As long as we have the UUID of the user I can pull back all the information for that user so we don't need anything user related but that. If the current logs don't give enough information about what the application state is at that given time then that's a wider conversation / PR. Logging is hard.

remybach:
I'm a bit worried about doing too much of a refactor in terms of that with the application logging as it seems to me to border on being feature work.

As for now, this PR _should_ eliminate the PII leakage.

worker/sync/db-persist/spoor-publish.js
richardLuko: you are changing this bcause of leaked pwd, right?
if so, why not simply change that place and leave rest as it was

umbobabo:
@remybach Sounds we have consensus. Shall we then just keep the uuid and merge this as it is?

worker/sync/db-persist/spoor-publish.js
remybach: Because it's not just passwords that are leaking and using `n-mask-logger` will handle all the other cases as well (including places where we can't tell via the code that PII is leaking).

PR #145, opened 2019-07-03
@maxcbc/further log updates

maxcbc:
add route to controller logs
fix dodgy type check

### DIFF
--- a/server/controllers/backup.js
+++ b/server/controllers/backup.js
@@ -42,7 +42,10 @@ module.exports = exports = async (req, res, next) => {
 		});
 	}
 	catch(error) {
-		log.error({error});
+		log.error({
+			event: 'BACKUP_ERROR',
+			error
+		});
 
 		res.sendStatus(500);
 	}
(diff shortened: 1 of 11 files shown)


server/lib/enrich/index.js
```diff
@@ -7,7 +7,7 @@ module.exports = exports = function enrich(content, format) {
 		content.type = content.content_type;
 	}
 
-	if (content instanceof Object && !(content instanceof Error) && content.type in exports) {
+	if (contenta.constructor.name === 'Object' && !(content instanceof Error) && content.type in exports) {
```
leggsimon:
```suggestion
	if (content.constructor.name === 'Object' && !(content instanceof Error) && content.type in exports) {
```

server/controllers/backup.js
```diff
@@ -42,7 +42,10 @@ module.exports = exports = async (req, res, next) => {
 		});
 	}
 	catch(error) {
-		log.error({error});
+		log.error({
+			route: req.route.path,
```
leggsimon: These aren't easily searchable in the codebase though

maxcbc: OK will do again

leggsimon: Ideally, some sort of searchable string would be good, even if it’s something like the function name uppercased

maxcbc: REQUEST_COMPLETED_MATE

PR #169, opened 2019-10-28
Increasing the Threshold for 401 error spike alerts

andrewgeorgiou1981:
The 401 error spike alert goes off quite often - after looking into it, it seems that as there is low traffic to the service, even a single visit to https://www.ft.com/republishing/contract by a non-syndicated user will generate an alert.
I'm not sure what the level should be; however updating the proportion to 10% of all requests might be an improvement.

Screenshot 2019-10-28 at 11 34 23


adambraimbridge:
[Return 403s instead of 500s for unauthorised access to next-syn-list](https://trello.com/c/1ItX3gSP/1273-return-403s-instead-of-500s-for-unauthorised-access-to-next-syn-list)

PR #177, opened 2019-12-18
Removes ft-next-syndication-api-https

binaryberry:
There were two health check links under the HealthCheck header both of these pages point to the same health check so I have decided to remove ft-next-syndication-api-https

This health check page will then be deleted from biz-ops


binaryberry:
I've opened this PR on behalf of Courtney, who is unable to access her Github account. As a consequence I can't post a review on that PR so I'm using this message to say I hereby approve of it!

PR #224, opened 2020-08-03
feat: check user permissions for downloads

tolujimoh:
This PR adds user permission check before content can be downloaded

### DIFF
--- a/server/helpers/is-download-disabled.js
+++ b/server/helpers/is-download-disabled.js
@@ -1,0 +1,13 @@
+'use strict';
+
+module.exports = exports  = (item, user) => {
+	return [
+		item.type === 'package',
+		item.notAvailable === true,
+		item.canBeSyndicated === 'verify',
+		item.canBeSyndicated === 'withContributorPayment' && user.contributor_content !== true,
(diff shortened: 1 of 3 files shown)


server/helpers/is-download-disabled.js
```diff
@@ -0,0 +1,13 @@
...
+
+module.exports = exports  = (item, user) => {
+	return [
+		item.type === 'package',
+		item.notAvailable === true,
+		item.canBeSyndicated === 'verify',
+		item.canBeSyndicated === 'withContributorPayment' && user.contributor_content !== true,
+		item.canBeSyndicated === 'no',
+		!item.canBeSyndicated,
+		item.canDownload < 1
```
AniaMakes, line 11: A thought - you'll never get this number lower than 0, so why can't this be `item.canDownload === 0`

tolujimoh: That should be correct, however, I left it this way to be consistent with other syndication repos
https://github.com/Financial-Times/n-syndication/blob/master/src/js/modal-download.js#L125

PR #227, opened 2020-08-07
ACC-546 enhanced enrich step to create "rich article" type 

AniaMakes:
### Why is the change made here?

We want the content type to be correct at the point on saving. This is happening in `server/controllers/save-by-content-id.js`.
That controller gets their `content` by `await getContentById()` (line 34)

In `server/lib/get-content-by-id.js` the content is fetched from Elastic Search (line 33 `await esClient.get()`).
In the same file, the content fetched in enriched (line 48 `enrich()`)

The enrich main file `server/lib/enrich/index.js` enriches the content by passing it into the relevant enrich helper. 

I added the rich article enricher after the base enricher, so that as a fail safe there'll always be a type of article.

### DIFF
--- a/server/lib/enrich/article.js
+++ b/server/lib/enrich/article.js
@@ -25,6 +25,10 @@ module.exports = exports = function article(content, format) {
 		content.content_type = CONTENT_TYPE_ALIAS[content.type] || content.type;
 	}
 
+	if (content.contentStats && content.contentStats.graphics && content.contentStats.graphics > 0){
+		content.content_type = 'rich article'
+	}
+
 	content.extension = DOWNLOAD_ARTICLE_FORMATS[format] || 'docx';
 
 	if (content.body && !content.bodyHTML) {


server/lib/enrich/article.js
```diff
@@ -25,6 +25,10 @@ module.exports = exports = function article(content, format) {
 		content.content_type = CONTENT_TYPE_ALIAS[content.type] || content.type;
 	}
 
+	if (content.contentStats && content.contentStats.graphics && content.contentStats.graphics > 0){
+		content.content_type = 'rich article'
+	}
+
```
tolujimoh, line 31: This could introduce data integrity issues and it would break other parts of `next-syndication-api` that rely on `content_type === 'article'`. 

This logic would better reside in the [export.js](https://github.com/Financial-Times/next-syndication-api/blob/master/server/controllers/export.js#L32-L34) controller (At the point of creating the csv file).

server/lib/enrich/article.js
```diff
@@ -25,6 +25,10 @@ module.exports = exports = function article(content, format) {
 		content.content_type = CONTENT_TYPE_ALIAS[content.type] || content.type;
 	}
 
+	if (content.contentStats && content.contentStats.graphics && content.contentStats.graphics > 0){
+		content.content_type = 'rich article'
```
rosario, line 29: Better with a constant value instead of a string `rich article` 

I'm not familiar with project but it seems from the live above there's CONTENT_TYPE_ALIAS lookup table?

server/lib/enrich/article.js
AniaMakes: Yes, it would be nice, but what that export does is reads the database and copy-pastes it into the csv file. 
Because of that we need to make sure that either 1) the database saves the count of rich content in a separate column or 2) (preferred, in my opinion) when we are sending stuff to the database, it can have a type of `rich article`

AniaMakes: Either way we'd need to change the database schema (the content types are an `enum` so that would need to be updated). The alternative - would be to ping Elastic search for each article we're trying to export to figure out if it's a rich article - that feels more "energy intensive".

AniaMakes:
This is the result of a 🧠  malfunction. This should have been a draft to start with. My bad. I'm closing this and will open a draft / actual PR when there's something that's actually worth looking at 😅

Issue #234, opened 2020-09-04
Enriching Audio / Podcast content does not work with 2020 ES results. 


AniaMakes:
As part of ACC-622 I was updating fixtures. They were created in 2017, and the ES results have been enriched multiple times since. 

When I try to update the `98b46b5f-17d3-40c2-8eaa-082df70c5f01` fixture ("The Economics of Migration" podcast"), the tests start failing. 

in `test/fixtures/content`, if I update...

 **JUST items.json:**
 1) server/controllers/resolve.spec.js return an Array of distinct content items for every content ID it can find: [ it doesn’t match, but content the same??]
2) server/lib/get-content.spec.js test return an Array of content items for every content ID it can find: [got 5, expected 6]

**BOTH items.json and the ES fixture file**
```
/Users/anna.bebb/code/next-syndication-api/test/server/lib/download/index.spec.js:43
		const className = `${content.type.charAt(0).toUpperCase()}${content.type.substring(1)}Download`;
		                             ^
TypeError: Cannot read property 'type' of null
```

**JUST ES fixture file**
```
/Users/anna.bebb/code/next-syndication-api/test/server/lib/download/index.spec.js:43
		const className = `${content.type.charAt(0).toUpperCase()}${content.type.substring(1)}Download`;
		                             ^

TypeError: Cannot read property 'type' of null
```

Having rubber-ducked this with @thurstontye , we discovered that the 2017 Elastic Search result for podcast was ` "type": "podcast",` (line 19), and the 2020 version is ` "type": "audio", "subtype": "podcast",`. As far as we can figure out, the enriching logic is built around the `type` being `podcast`, so with the changed structure this fails. 

**It is highly likely this is currently happening on prod.** However, we think this actually isn't an immediate problem because podcasts are free content (you can listen to them on any podcast app, you don't need to pay the FT subscription) and therefore not subject to syndication in the first place? Either way, this investigation is out of scope for ACC-622 and I can proceed without updating this particular fixture, but it feels like this needs fixing in the longer term, or deleting altogether if this sort of content doesn't fall under syndication rules?

PR #235, opened 2020-09-07
ACC-622 Enhance response to include info about Graphic syndication

AniaMakes:
## Ticket
https://financialtimes.atlassian.net/browse/ACC-622

## Things done in this PR
- updated some fixtures to their 2020 state (fixtures are used for mocks in tests and have been stuck in their 2017 version). Not all of the fixtures have been updated (see #234 ) because it was out of scope for this ticket. 
- the endpoint now returns `hasGraphics` boolean and `canAllGraphicsBeSyndicated` boolean *for article only*

## Shortfalls
Currently we don't have an article that has `canAllGraphicsBeSyndicated` `true`, and I'm not sure what's the practicality of having a test for something that is different in prod.

### DIFF
--- a/server/lib/enrich/article.js
+++ b/server/lib/enrich/article.js
@@ -33,6 +33,11 @@ module.exports = exports = function article(content, format) {
 
 	content.hasGraphics = Boolean(content.contentStats && content.contentStats.graphics);
 
+	const countOfGraphics = content.contentStats && content.contentStats.graphics && content.contentStats.graphics;
+	const countOfSharableGraphics = content.embeds ? content.embeds.filter(embed => embed && embed.type.endsWith('Graphic')).filter(embed => embed.canBeSyndicated === 'yes').lenght : 0;
+
+	content.canAllGraphicsBeSyndicated = countOfGraphics > 0 ? countOfSharableGraphics < countOfGraphics : null;
+
 	if (content.bodyHTML) {
 		content.document = formatArticleXML(`<body>${content.bodyHTML}</body>`);
 
--- a/server/lib/enrich/article.js
+++ b/server/lib/enrich/article.js
@@ -39,7 +39,7 @@ module.exports = exports = function article(content, format) {
 	// can't be, the answer to this is 'no'
 	const atLeastOneGraphicCantBeShared = content.embeds && content.embeds.filter(embed => embed && embed.type.endsWith('Graphic')).some(item => item.canBeSyndicated !== 'yes');
 
-	content.canAllGraphicsBeSyndicated = Boolean(atLeastOneGraphicCantBeShared);
+	content.canAllGraphicsBeSyndicated = !atLeastOneGraphicCantBeShared;
 
 	if (content.bodyHTML) {
 		content.document = formatArticleXML(`<body>${content.bodyHTML}</body>`);
--- a/server/lib/enrich/article.js
+++ b/server/lib/enrich/article.js
@@ -36,7 +36,7 @@ module.exports = exports = function article(content, format) {
 	const countOfGraphics = content.contentStats && content.contentStats.graphics && content.contentStats.graphics;
 	const countOfSharableGraphics = content.embeds ? content.embeds.filter(embed => embed && embed.type.endsWith('Graphic')).filter(embed => embed.canBeSyndicated === 'yes').lenght : 0;
 
-	content.canAllGraphicsBeSyndicated = countOfGraphics > 0 ? countOfSharableGraphics < countOfGraphics : null;
+	content.canAllGraphicsBeSyndicated = countOfGraphics > 0 ? countOfSharableGraphics < countOfGraphics : false;
 
 	if (content.bodyHTML) {
 		content.document = formatArticleXML(`<body>${content.bodyHTML}</body>`);
--- a/server/lib/enrich/article.js
+++ b/server/lib/enrich/article.js
@@ -33,10 +33,13 @@ module.exports = exports = function article(content, format) {
 
 	content.hasGraphics = Boolean(content.contentStats && content.contentStats.graphics);
 
-	const countOfGraphics = content.contentStats && content.contentStats.graphics;
-	const countOfSharableGraphics = content.embeds ? content.embeds.filter(embed => embed && embed.type.endsWith('Graphic') && embed.canBeSyndicated === 'yes').length : 0;
+	// Currently embeds can have more than one item for each picture - for different screen sizes
+	// We are assuming that canBeSyndicated is the same for all sizes of a picture
+	// We are asking if ALL Graphics can be syndicated, which means as long as at least one item
+	// can't be, the answer to this is 'no'
+	const atLeastOneGraphicCantBeShared = content.embeds && content.embeds.filter(embed => embed && embed.type.endsWith('Graphic')).some(item => item.canBeSyndicated !== 'yes');
 
-	content.canAllGraphicsBeSyndicated = countOfGraphics > 0 ? countOfSharableGraphics < countOfGraphics : false;
+	content.canAllGraphicsBeSyndicated = Boolean(atLeastOneGraphicCantBeShared);
 
 	if (content.bodyHTML) {
 		content.document = formatArticleXML(`<body>${content.bodyHTML}</body>`);
--- a/server/lib/enrich/article.js
+++ b/server/lib/enrich/article.js
@@ -34,7 +34,7 @@ module.exports = exports = function article(content, format) {
 	content.hasGraphics = Boolean(content.contentStats && content.contentStats.graphics);
 
 	const countOfGraphics = content.contentStats && content.contentStats.graphics && content.contentStats.graphics;
-	const countOfSharableGraphics = content.embeds ? content.embeds.filter(embed => embed && embed.type.endsWith('Graphic')).filter(embed => embed.canBeSyndicated === 'yes').lenght : 0;
+	const countOfSharableGraphics = content.embeds ? content.embeds.filter(embed => embed && embed.type.endsWith('Graphic')).filter(embed => embed.canBeSyndicated === 'yes').length : 0;
 
 	content.canAllGraphicsBeSyndicated = countOfGraphics > 0 ? countOfSharableGraphics < countOfGraphics : false;
 
--- a/server/lib/enrich/article.js
+++ b/server/lib/enrich/article.js
@@ -33,7 +33,7 @@ module.exports = exports = function article(content, format) {
 
 	content.hasGraphics = Boolean(content.contentStats && content.contentStats.graphics);
 
-	const countOfGraphics = content.contentStats && content.contentStats.graphics && content.contentStats.graphics;
+	const countOfGraphics = content.contentStats && content.contentStats.graphics;
 	const countOfSharableGraphics = content.embeds ? content.embeds.filter(embed => embed && embed.type.endsWith('Graphic')).filter(embed => embed.canBeSyndicated === 'yes').length : 0;
 
 	content.canAllGraphicsBeSyndicated = countOfGraphics > 0 ? countOfSharableGraphics < countOfGraphics : false;
--- a/server/lib/enrich/article.js
+++ b/server/lib/enrich/article.js
@@ -34,7 +34,7 @@ module.exports = exports = function article(content, format) {
 	content.hasGraphics = Boolean(content.contentStats && content.contentStats.graphics);
 
 	const countOfGraphics = content.contentStats && content.contentStats.graphics;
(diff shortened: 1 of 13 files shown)


server/lib/enrich/article.js
```diff
@@ -31,6 +31,13 @@ module.exports = exports = function article(content, format) {
 		content.bodyHTML = content.body;
 	}
 
+	content.hasGraphics = content.contentStats && content.contentStats.graphics && content.contentStats.graphics > 0 ? true : false;
```
AniaMakes, line 34: If you're wondering why the `? true : false`, it's because `content.contentStats && content.contentStats.graphics && content.contentStats.graphics > 0` returns 0 when `graphics: 0`. Would `!!` be a solution here?

server/lib/enrich/article.js
```diff
@@ -31,6 +31,13 @@ module.exports = exports = function article(content, format) {
...
+
+	const countOfGraphics = content.contentStats && content.contentStats.graphics && content.contentStats.graphics;
+	const countOfSharableGraphics = content.embeds ? content.embeds.filter(embed => embed && embed.type.endsWith('Graphic')).filter(embed => embed.canBeSyndicated === 'yes').length : 0;
+
+	content.canAllGraphicsBeSyndicated = countOfGraphics > 0 ? countOfSharableGraphics < countOfGraphics : false;
```
AniaMakes, line 39: Decided to go with `false` rather than `null` because that way I can have `expect a boolean` in tests.

server/lib/syndicate-content.js
```diff
@@ -70,5 +70,10 @@ function tidy(item, includeBody) {
 		delete item.fileName;
 	}
 
+	if (item.hasGraphics === undefined){
```
AniaMakes, line 73: This bit removes the keys and their `undefined` values from non-article types.

server/lib/enrich/article.js
rosario: no need to use a ternary there:

```> content = {contentStats :{graphics: 0}}
{ contentStats: { graphics: 0 } }
> content.contentStats && content.contentStats.graphics && content.contentStats.graphics > 0
0
> content.contentStats && content.contentStats.graphics > 0
false
```
It returns `0` because `somethingTruthy && 0` is `0`


This would work
```
content.contentStats && content.contentStats.graphics > 0
```

server/lib/enrich/article.js
```diff
@@ -31,6 +31,13 @@ module.exports = exports = function article(content, format) {
 		content.bodyHTML = content.body;
 	}
 
+	content.hasGraphics = content.contentStats && content.contentStats.graphics && content.contentStats.graphics > 0 ? true : false;
+
+	const countOfGraphics = content.contentStats && content.contentStats.graphics && content.contentStats.graphics;
```
rosario, line 36: there's a duplicate `content.contentStats.graphics`

server/lib/enrich/article.js
AniaMakes: So if there's no `graphics` key, that would be `false`, right?

server/lib/enrich/article.js
```diff
@@ -31,6 +31,13 @@ module.exports = exports = function article(content, format) {
...
+	const countOfSharableGraphics = content.embeds ? content.embeds.filter(embed => embed && embed.type.endsWith('Graphic')).filter(embed => embed.canBeSyndicated === 'yes').length : 0;
```
rosario, line 37: Instead of using `filter` twice, you could use one filter and have both conditions. 

Although the performance is not really an issue (the array is scanned twice with two filters) I think it's more readable to have the "predicate" of the filter to be one in this case. 

When will I use two filters? If the predicates where complicated enough to deserve to be separated

server/lib/enrich/article.js
rosario: You can test it easily in your node console :) 

```
> content.contentStats && content.contentStats.graphics > 0
false
> content.contentStats && content.contentStats.SOMETHINGELSE > 0
false
> undefined > 0
false
```

server/lib/enrich/article.js
```diff
@@ -31,6 +31,13 @@ module.exports = exports = function article(content, format) {
...
+
+	content.canAllGraphicsBeSyndicated = countOfGraphics > 0 ? countOfSharableGraphics < countOfGraphics : false;
+
```
tolujimoh, line 40: The `embeds` array contains all the images and graphics in all their different sizes. However, only the original (default size) images are used for syndication.

Should the graphics included in the content document be the source of truth for graphics in the article instead of the embeds array? 

the embeds array would then be used as a map to check whether the graphics can be syndicated or not.

server/lib/syndicate-content.js
```diff
@@ -70,5 +70,10 @@ function tidy(item, includeBody) {
...
+		delete item.canAllGraphicsBeSyndicated;
+		delete item.hasGraphics;
+	}
+
```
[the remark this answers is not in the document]
AniaMakes: Ah, you'd think so, wouldn't you. That would make sense! But in `server/lib/syndicate-content.js` there's this spaghetti where `RESOLVE_PROPERTIES` is a list of properties, set up in `lib/resolve.js` that then has a forEach to assign the property to that key. We could make a special case of 'rich article' but it's simpler to set it up and then clean it up.

server/lib/enrich/article.js
AniaMakes: But that's what's there, with `content.hasGraphics` relying on `contentStats.graphics` count? (line 34)?

Then you have `countOfGraphics` (once again based on the `contentStats`). And `countOfSharableGraphics` which runs through the `embeds` to see how many can be syndicated. 

It feels like you know something I don't - I'm just using an example ES article to figure this out 😅

server/lib/enrich/article.js
AniaMakes: It still feels like code is sometimes pure magic 😂

rosario (approved):

server/lib/syndicate-content.js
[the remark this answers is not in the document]
tolujimoh: The inclusion of these files [server/lib/resolve/canAllGraphicsBeSyndicated.js](https://github.com/Financial-Times/next-syndication-api/pull/235/files#diff-892e3de415be564e9d8cc859f3e6022a) and [server/lib/resolve/hasGraphics.js](https://github.com/Financial-Times/next-syndication-api/pull/235/files#diff-f97b3555f2dab9fdaa4445c93df214a4) is what adds the properties to non-articles. They also don't appear to serve any additional purpose for articles as `hasGraphics` and `canAllGraphicsBeSyndicated` properties are all ready included in the item object from the enrichment (or is there something I missing). 

It seems like those files are not needed

server/lib/enrich/article.js
tolujimoh: contentStats.graphics is gotten from [counting the number of graphics](https://github.com/Financial-Times/next-es-interface/blob/master/server/transformers/plugins/article/content-stats.js#L25) in the content document. embeds contains all graphics and their different sizes.

For example, the count for `contentStats.graphics` can be `1` and the count of graphics object in embeds could be `2` (graphics object for normal screens, graphics object for mobile screens). 

Therefore, using that to count the number of sharable graphics can be inaccurate. 

As we do not have any identifier for graphics object for normal screen (which is the graphics size we are using for syndication) in the embeds array.

We can
1) Add a way to identify graphics objects for normal screens in the embeds array. That would require work on [next-es-interface](https://github.com/Financial-Times/next-es-interface/blob/master/server/models/embeds/index.js)
2) Get the graphics included in the content document, and look up the graphics object in the embeds array to check the `canBeSyndicated` property.

server/lib/resolve/index.js
```diff
@@ -1,10 +1,12 @@
 'use strict';
 
 exports.id = require('./id');
+exports.canAllGraphicsBeSyndicated = require('./canAllGraphicsBeSyndicated');
 exports.canDownload = require('./canDownload');
 exports.canBeSyndicated = require('./canBeSyndicated');
 exports.downloaded = require('./downloaded');
 exports.embargoPeriod = require('./embargoPeriod');
+exports.hasGraphics = require('./hasGraphics');
```
umbobabo, line 9: Are you sure that these files are necessary? Considering that you then remove the properties [here](https://github.com/Financial-Times/next-syndication-api/pull/235/files#diff-e2babe4577db9910d30c6c07532ca665R73) and that what you your code execute during the enrichment and the enrichment seems to specific only for articles.

tolujimoh: Looks like it is necessary. The resolve controller passes an empty object to `item` and the enriched content to `src`.
https://github.com/Financial-Times/next-syndication-api/blob/master/server/controllers/resolve.js#L47-L53

If the property isn't included in the folder, it doesn't get set in the item object.
https://github.com/Financial-Times/next-syndication-api/blob/master/server/lib/syndicate-content.js#L40

AniaMakes: Yeah, it's really odd way to do it, but I'm guessing whoever built it has a reason to do so?

umbobabo: They do that cause the same function is [used here](https://github.com/Financial-Times/next-syndication-api/blob/42bb5ac58119c40b34a1598632debb2909f8d608/server/controllers/translations.js#L132) but passing the original item to prepare the syndicate object for the Spanish content.

Let's keep in mind about the ES part. Do we need to do anything about that?

server/lib/enrich/article.js
```diff
@@ -33,6 +33,14 @@ module.exports = exports = function article(content, format) {
 
 	content.hasGraphics = Boolean(content.contentStats && content.contentStats.graphics);
 
+	// Currently embeds can have more than one item for each picture - for different screen sizes
+	// We are assuming that canBeSyndicated is the same for all sizes of a picture
+	// We are asking if ALL Graphics can be syndicated, which means as long as at least one item
+	// can't be, the answer to this is 'no'
+	const atLeastOneGraphicCantBeShared = content.embeds && content.embeds.filter(embed => embed && embed.type.endsWith('Graphic')).some(item => item.canBeSyndicated !== 'yes');
+
+	content.canAllGraphicsBeSyndicated = Boolean(atLeastOneGraphicCantBeShared);
```
alexnaish, line 42: This is confusing for me - the content property is a positive assertion but then its being converted to a boolean of a negative assertion?

server/lib/enrich/article.js
```diff
@@ -33,6 +33,14 @@ module.exports = exports = function article(content, format) {
 
 	content.hasGraphics = Boolean(content.contentStats && content.contentStats.graphics);
 
+	// Currently embeds can have more than one item for each picture - for different screen sizes
+	// We are assuming that canBeSyndicated is the same for all sizes of a picture
+	// We are asking if ALL Graphics can be syndicated, which means as long as at least one item
+	// can't be, the answer to this is 'no'
+	const atLeastOneGraphicCantBeShared = content.embeds && content.embeds.filter(embed => embed && embed.type.endsWith('Graphic')).some(item => item.canBeSyndicated !== 'yes');
```
alexnaish, line 40: What are the possible values of each `embed`? I can't see an example in the fixtures.

server/lib/enrich/article.js
tolujimoh, line 42:
```suggestion
	content.canAllGraphicsBeSyndicated = !atLeastOneGraphicCantBeShared;
```
If at least one graphic can't be syndicated, then all graphics can't be syndicated.

Likewise, if there is no graphics that can't be syndicated, then all graphics can be syndicated.

server/lib/enrich/article.js
tolujimoh, line 40:
```suggestion
	const atLeastOneGraphicCantBeShared = content.embeds && content.embeds.some(embed => embed && embed.type.endsWith('Graphic') && embed.canBeSyndicated !== 'yes');
```
Can be refactored to.

server/lib/enrich/article.js
AniaMakes: Well spotted - confused myself with the double negatives and all 😅

AniaMakes: My bad. Got confused with double negatives 😅

server/lib/enrich/article.js
AniaMakes: If you don't mind, I find the staggered `filter - some` much easier to parse - I don't know how others feel about this.

tolujimoh: Also, 

Since the type is defined as `http://www.ft.com/ontology/content/Graphic`.
We can use equality for comparison instead of `endsWith('Graphic')`

Maybe personal preference

AniaMakes: https://github.com/Financial-Times/next-syndication-api/pull/235/files#diff-cda288f59227ef1098aee289e17d466eR48

tolujimoh (approved):

PR #237, opened 2020-09-08
fix: Inc 697 - Syndication users unable to download paid for content

tolujimoh:
Users with a licence that can download paid content should have contributor_content as true.

However, contributor_content should have be gotten from res.local.contract that is defined [here](https://github.com/Financial-Times/next-syndication-api/blob/master/server/middleware/get-contract-by-id-from-param.js#L12) not res.local.user.


alexnaish (approved):

jkerr321 (approved):
🙌

PR #243, opened 2020-09-16
feat: add has graphics when enriching

tolujimoh:
This PR adds `hasGraphics` property when enriching content

### DIFF
--- a/server/lib/enrich/article.js
+++ b/server/lib/enrich/article.js
@@ -26,10 +26,12 @@ module.exports = exports = function article(content, format) {
 	}
 
 	content.extension = DOWNLOAD_ARTICLE_FORMATS[format] || 'docx';
-
+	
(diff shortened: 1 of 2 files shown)


server/lib/enrich/article.js
```diff
@@ -31,6 +31,8 @@ module.exports = exports = function article(content, format) {
 		content.bodyHTML = content.body;
 	}
 
+	content.hasGraphics = content.contentStats && content.contentStats.graphics > 0;
+
```
rosario, line 35: if `contentStats` is undefined you'll get undefined:
```
> content = {}
{}
> content.contentStats && content.contentStats.graphics > 0
undefined
```

tolujimoh: Good spot! 👍

rosario (approved):

PR #246, opened 2020-09-28
ACC-545 Contract Details page update

thurstontye:
Updates contract.content_allowed to contain either Article or Rich Article.

This is a draft PR until an example of a saleforce 'rich article' account it's not clear if the changes made here are the correct thing to do.


server/helpers/has-rich-article.js
```diff
@@ -0,0 +1,5 @@
+'use strict';
+
+module.exports = exports = (prodList= []) => {
+	return (prodList.length >= 1) ? prodList.some(product => product.code === 'S2'): false;
+};
```
rosario, line 5: My 2p, if this function is not used anywhere else then it doesn't need to be into a separate file itself.
In particular if it's such simple one line function.

rosario: A note about `S2`, in master it's defined as a constant now:

https://github.com/Financial-Times/next-syndication-api/blob/master/config/default.yaml#L246

thurstontye: @rosario just realised there is a bunch of changes in here that are no longer relevant. Hold fire on reviewing this I'm going to tidy up the PR

thurstontye:
Out of date changes new PR on it's way

PR #254, opened 2020-10-06
ACC-558 Added spike health check for syndication-dl

AniaMakes:
Currently next-syndication-dl doesn't have any healthchecks, so we are not alerted about any spikes in 404s (which are the symptoms of issues such as this one). This healtcheck will help to spot issues ourselves rather than waiting on Customer Care to report this from users.

**Why is this not a "number of successes vs number of failures"?** The number of `200` responses isn't high enough on regular basis for the ratio to be a sensible measure.  

Reasoning for numbers:
severity - 1 - in the incident linked above it was stated that it ended up being a P2 but it should really have been a P1
threshold for 404 - 5 - it feels like 2 or 3 can happen organically every now and again, but anything over 5 would benefit from at least a glance what's happening
threshold for 403 - 1 - not sure about this one, we rarely get any (we had total of 2 last week, 3 in the past 30 days).

### DIFF
--- a/health/dl-error-spikes.js
+++ b/health/dl-error-spikes.js
@@ -1,31 +1,21 @@
 const nHealth = require('n-health');
 const samplePeriod = 10; // minutes
+const threshold = 5;
 
 // 3 and 5 are the positions of wildcards in the chain, 0 based
 // 6 is the node the count is based on, 0 based
-const metric403 = 'aliasByNode(sumSeriesWithWildcards(next.heroku.$app.*.express.* _{ GET, POST, PUT }.res.status.403.count, 3, 5), 6)';
-const metric404 = 'aliasByNode(sumSeriesWithWildcards(next.heroku.$app.*.express.* _{ GET, POST, PUT }.res.status.404.count, 3, 5), 6)';
-
+const metric = 'aliasByNode(sumSeriesWithWildcards(next.heroku.syndication-dl.*.express.* _{ GET, POST, PUT }.res.status_5xx.count, 3, 5), 6)';
 const panicGuide = 'Check the heroku logs for the app for any error messages (`heroku logs --app ft-next-syndication-dl --tail --num 100` - num being the number of lines to retrieve)'
 
-const services = [
-	{
-		id: 'syndication-download-403-forbidden-spike',
-		metric: metric403,
-		threshold: 1,
-		name: '403 rate for next-syndication-dl is acceptable',
-		businessImpact: 'Syndication users are experiencing issues downloading articles',
-		technicalSummary: `The count of 403 (Forbidden) responses for syndication download requests across all heroku dynos is higher than a threshold of 1 over the last ${samplePeriod} minutes`,
-	},
-	{
-		id: 'syndication-download-404-not-found-spike',
-		metric: metric404,
-		threshold: 5,
-		name: '404 rate for next-syndication-dl is acceptable',
-		businessImpact: 'Syndication users are experiencing issues downloading articles',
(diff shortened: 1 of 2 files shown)


health/dl-error-spikes.js
```diff
@@ -0,0 +1,31 @@
...
+
+const panicGuide = 'Check the heroku logs for the app for any error messages (`heroku logs --app ft-next-syndication-dl --tail --num 100` - num being the number of lines to retrieve)'
+
+const services = [
+	{
+		id: 'syndication-download-403-forbidden-spike',
+		metric: metric403,
+		threshold: 1,
+		name: '403 rate for next-syndication-dl is acceptable',
+		businessImpact: 'Syndication users are experiencing issues downloading articles',
```
AniaMakes, line 17: I'm not entirely sure what should be the business impact of 403s

health/dl-error-spikes.js
```diff
@@ -0,0 +1,21 @@
+const nHealth = require('n-health');
+const samplePeriod = 10; // minutes
+const threshold = 5;
+
+// 3 and 5 are the positions of wildcards in the chain, 0 based
+// 6 is the node the count is based on, 0 based
+const metric = 'aliasByNode(sumSeriesWithWildcards(next.heroku.syndication-dl.*.express.* _{ GET, POST, PUT }.res.status_5xx.count, 3, 5), 6)';
```
tolujimoh, line 7:
```suggestion
const metric = 'aliasByNode(sumSeriesWithWildcards(next.heroku.syndication-dl.*.express.* _{ GET, POST, PUT }.res.status_4xx.count, 3, 5), 6)';
```

Shouldn't it be `4xx`

AniaMakes: Yes, I was looking too hard at the example Umberto posted. D'oh.

tolujimoh (approved):
LGTM 👍 

However is there a way we can test the query
```
aliasByNode(sumSeriesWithWildcards(next.heroku.syndication-dl.*.express.* _{ GET, POST, PUT }.res.status_5xx.count, 3, 5), 6)
```

cc @umbobabo

tolujimoh (changes requested):
@AniaMakes We can try out and test the query in the explore tab on grafana dashboard.

The current query wouldn't work as `status_4xx` wouldn't match with any property on the res object.

Here is a link to the [correct query](https://grafana.ft.com/explore?orgId=1&left=%5B%22now-7d%22,%22now%22,%22graphitev2%22,%7B%22datasource%22:%22graphitev2%22,%22target%22:%22aliasByNode(sumSeriesWithWildcards(next.heroku.syndication-dl.*.express.*.res.status.4*.count,%203,%205),%206)%22,%22textEditor%22:false%7D,%7B%22mode%22:%22Metrics%22%7D,%7B%22ui%22:%5Btrue,true,true,%22none%22%5D%7D%5D)

health/dl-error-spikes.js
```diff
@@ -0,0 +1,21 @@
+const nHealth = require('n-health');
+const samplePeriod = 30; // minutes
+const threshold = 5;
+
+// 3 and 5 are the positions of wildcards in the chain, 0 based
+// 6 is the node the count is based on, 0 based
+const metric = `summarize(sumSeries(next.heroku.syndication-dl.*.express.*.res.status.*.count), "${samplePeriod}min", "sum", true)`;
```
tolujimoh, line 7:
```suggestion
const metric = `summarize(sumSeries(next.heroku.syndication-dl.*.express.*.res.status.4*.count), "${samplePeriod}min", "sum", true)`;
```

tolujimoh (approved):

PR #255, opened 2020-10-06
refactor: use jsdom instead of xmldom

tolujimoh:
To select elements in the DOM document by query (e.g. `img[image_type=graphic]`) we need to use [querySelectorAll](https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelectorAll) method.

[xmldom](https://github.com/jindw/xmldom) which is used by syndication api to [parse the HTML string to a DOM Document](https://github.com/Financial-Times/next-syndication-api/blob/88e39c42aeffc8f3a95e76c9b8c58793140dd609/server/lib/format-article-xml.js#L12) uses [W3C DOM Level 2](https://www.w3.org/TR/2000/REC-DOM-Level-2-Core-20001113/core.html) which doesn't have querySelectorAll method in the Document Object.

[jsdom](https://github.com/jsdom/jsdom) uses [WHATWG DOM and HTML Standards](https://dom.spec.whatwg.org/) which has querySelectorAll method in the Document Object.

As such, this PR replaces xmldom with jsdom.


server/lib/decorate-article.js
```diff
@@ -5,15 +5,15 @@ const path = require('path');
...
 const moment = require('moment');
-const { DOMParser } = require('xmldom');
+const { JSDOM } = require('jsdom');
 
 const Handlebars = handlebars();
 
 const BASE_PATH = path.dirname(path.relative(process.cwd(), __dirname));
-const HD = Handlebars.compile(fs.readFileSync(path.resolve(BASE_PATH, './views/partial/article_metadata_hd.html.hbs'), 'utf8'), { noEscape: true });
-const FT = Handlebars.compile(fs.readFileSync(path.resolve(BASE_PATH, './views/partial/article_metadata_ft.html.hbs'), 'utf8'), { noEscape: true });
+const headerTemplate = Handlebars.compile(fs.readFileSync(path.resolve(BASE_PATH, './views/partial/article_metadata_hd.html.hbs'), 'utf8'), { noEscape: true });
```
thurstontye, line 13: much better, easier to read

server/lib/enrich/article.js
```diff
@@ -54,12 +54,12 @@ module.exports = exports = function article(content, format) {
 
 		content.document = decorateArticle(content.document, content);
 
-		content.bodyHTML__CLEAN = content.document.toString();
+		content.bodyHTML__CLEAN = content.document.body.outerHTML;
```
thurstontye, line 57: Is this definately equivalent ot document.toString()?

server/lib/decorate-article.js
```diff
@@ -5,15 +5,15 @@ const path = require('path');
...
+const footerTemplate = Handlebars.compile(fs.readFileSync(path.resolve(BASE_PATH, './views/partial/article_metadata_ft.html.hbs'), 'utf8'), { noEscape: true });
```
AniaMakes, line 14: Really good idea to update the names, makes it much clearer! 🍪

server/lib/enrich/article.js
```diff
@@ -54,12 +54,12 @@ module.exports = exports = function article(content, format) {
...
 
 		// we need to strip all formatting — leaving only paragraphs — and pass this to pandoc for plain text
 		// otherwise it will uppercase the whole article title and anything bold, as well as leave other weird
 		// formatting in the text file
-		content.bodyHTML__PLAIN = toPlainText(content.document.toString());
+		content.bodyHTML__PLAIN = toPlainText(content.document.body.outerHTML);
```
thurstontye, line 62: Why not do `content.bodyHTML__PLAIN = toPlainText(content.bodyHTML__CLEAN);`?

thurstontye (approved):
LGTM 👍

server/lib/get-word-count.js
```diff
@@ -2,8 +2,8 @@
 
 const RE_VALID_ONE_CHARS = /^[IAa0-9]$/;
```
AniaMakes, line 4: Given that in that in another PR your replaces the home brew function, is this file actually doing anything?

AniaMakes (commented):
This looks OK to me, but this is on "reading this, it makes sense". I guess as long as tests pass, then it works? 😅

server/lib/enrich/article.js
tolujimoh, line 62:
```suggestion
		content.bodyHTML__PLAIN = toPlainText(content.bodyHTML__CLEAN);
```

PR #256, opened 2020-10-07
chore: add stub for contract with graphics syndication

tolujimoh:
The PR adds the stub for a contract with graphics syndication


rosario (approved):

stubs/FTS-14076678.json
```diff
@@ -0,0 +1,30 @@
+{
+    "videoLimit": null,
```
thurstontye, line 2: No video assets lncluded so do we need this property? Or should we have video and podcasts on this contract too so it is similar to 'CA-00001558' but set up for rich articles

stubs/FTS-14076678.json
```diff
@@ -0,0 +1,30 @@
...
+    "success": true,
+    "startDate": "2020-04-28",
+    "podcastLimit": null,
```
thurstontye, line 5: same as with `videoLimit`

stubs/FTS-14076678.json
tolujimoh: We would need to update the assets from salesforce to include video and podcast

PR #259, opened 2020-10-14
Feat/testing update column csv

rosario:
Using `S2` product code to identify Graphic Syndications users and send the correct event with `has_graphics`

Reference: https://financialtimes.atlassian.net/browse/ACC-546?focusedCommentId=469784

Note: This PR will merge into @tolujimoh work on `feat/update-content-type-column-on-csv`


tolujimoh (approved):
Looks good. 👍

tolujimoh (commented):
Are we able to test the event received by the Message queue and subsequently sent to the `db-persist` sync job?

PR #262, opened 2020-10-21
ACC-545 rich article access

thurstontye:
Adds `rich_articles:true` to the user status 'allowed' object as a flag for client consumer of the endpoint to identify those with graphic syndication access.

NB: This takes a slightly different approach to ACC-546 to obtain the 'S2' product code. The reason for this deviation is the property `res.locals.hasGraphicSyndication` does not get set if `res.locals.EXPEDITED_USER_AUTH= true`. Given the product codes are availble in  `res.locals.licence.products` set in `getSyndicationLicenceForUser` which is called a  little later in the middleware than `isSyndicationUser` but importantly is not bypassed by `EXPEDITED_USER_AUTH` being set I think that may be a better place to do that. Change here accomodates both but I think it would be ok to just do the check in `getSyndicationLicenceForUser`. I'm fairly sure licence is source of truth for user product codes so I think it's safe to do so.

### DIFF
--- a/config/default.yaml
+++ b/config/default.yaml
@@ -10,7 +10,6 @@
     FT Article: Articles
     Podcast: Podcasts
     Video: Videos
-    FT Rich Article: Rich Articles
 
   BASE_URI_FT_API: https://api.ft.com
 
--- a/config/default.yaml
+++ b/config/default.yaml
@@ -10,6 +10,7 @@
     FT Article: Articles
     Podcast: Podcasts
     Video: Videos
+    FT Rich Article: Rich Articles
 
   BASE_URI_FT_API: https://api.ft.com
 
--- a/server/controllers/user-status.js
+++ b/server/controllers/user-status.js
@@ -7,7 +7,7 @@ const log = require('../lib/logger');
 const { FEATURE_FLAGS } = require('config');
 
 const flagIsOn = require('../helpers/flag-is-on');
-const hasRichArticleAccess = require('../helpers/has-rich-article');
+//const hasRichArticleAccess = require('../helpers/has-rich-article');
 
 const PACKAGE = require(path.resolve('./package.json'));
 
@@ -22,14 +22,12 @@ module.exports = exports = async (req, res, next) => {
 			isNewSyndicationUser,
 			licence,
 			syndication_contract,
-			user
+			user,
+			hasGraphicSyndication,
 		} } = res;
 
-
-		const rich_articles = hasRichArticleAccess(licence.products);
-
-		if(allowed && rich_articles) {
-			allowed.rich_articles =  rich_articles;
+		if(hasGraphicSyndication) {
+			allowed.rich_articles = hasGraphicSyndication;
 		}
 
 		const userStatus = Object.assign({
--- a/server/controllers/user-status.js
+++ b/server/controllers/user-status.js
@@ -14,8 +14,6 @@ const PACKAGE = require(path.resolve('./package.json'));
 module.exports = exports = async (req, res, next) => {
 	try {
 		res.status(200);
-		console.log('🤩');
-		console.log(res.locals.hasGraphicSyndication);
 
 		const { locals: {
 			allowed,
--- a/server/controllers/user-status.js
+++ b/server/controllers/user-status.js
@@ -7,7 +7,6 @@ const log = require('../lib/logger');
 const { FEATURE_FLAGS } = require('config');
 
 const flagIsOn = require('../helpers/flag-is-on');
-//const hasRichArticleAccess = require('../helpers/has-rich-article');
 
 const PACKAGE = require(path.resolve('./package.json'));
 
--- a/server/controllers/user-status.js
+++ b/server/controllers/user-status.js
@@ -21,13 +21,8 @@ module.exports = exports = async (req, res, next) => {
 			isNewSyndicationUser,
 			licence,
 			syndication_contract,
-			user,
-			hasGraphicSyndication,
+			user
 		} } = res;
-
-		if(hasGraphicSyndication) {
-			allowed.rich_articles = hasGraphicSyndication;
-		}
 
 		const userStatus = Object.assign({
 			app: {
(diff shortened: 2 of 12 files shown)


thurstontye:
@rosario @tolujimoh 👆 Alternatives to my suggestion would be to persist in the DB, which I think is overkill or making another request to the session service here https://github.com/Financial-Times/next-syndication-api/blob/48abdc0b0f6c592cf242e033bb3fc6fca1376091/server/middleware/expedite-user-auth.js#L23

server/controllers/user-status.js
```diff
@@ -7,6 +7,7 @@ const log = require('../lib/logger');
 const { FEATURE_FLAGS } = require('config');
 
 const flagIsOn = require('../helpers/flag-is-on');
+//const hasRichArticleAccess = require('../helpers/has-rich-article');
```
rosario, line 10: forgot to remove this

server/helpers/has-rich-article.js
```diff
@@ -0,0 +1,4 @@
+'use strict';
+module.exports = exports = (prodList= []) => {
+	return (prodList.length >= 1) ? prodList.some(product => product.code === 'S2'): false;
+};
```
rosario, line 4: My 2p, there's no need to put this one liner in a separated file, it's only used once. 

If it needs to be in separated file, then it needs test

server/middleware/get-syndication-licence-for-user.js
```diff
@@ -54,6 +56,11 @@ module.exports = exports = async (req, res, next) => {
 
 		res.locals.licence = syndicationLicence;
 
+		if(!res.locals.hasGraphicSyndication) {
+			res.locals.hasGraphicSyndication = hasRichArticleAccess(syndicationLicence.products);
+		}
```
rosario, line 61: We'd need a test for this middleware to make sure we set `hasGraphicSyndication` in req.locals

thurstontye: 👍

server/helpers/has-rich-article.js
thurstontye: Tests added. Done as a helper as I was thining this might be used on different routes as an when but now it's in the middleware perhaps not.

server/helpers/has-graphics-access.js
```diff
@@ -0,0 +1,6 @@
+'use strict';
+
+const {GRAPHIC_SYNDICATION_PRODUCT_CODE} = require('config');
+module.exports = exports = (prodList= []) => {
+	return (prodList.length >= 1) ? prodList.some(product => product.code === GRAPHIC_SYNDICATION_PRODUCT_CODE): false;
```
rosario, line 5: Sorry for late comment, I only spotted it now:

There's no need to have a ternary operator, `some` applied to empty array will return false
```
> [].some(a => a)
false
```

thurstontye: of course 🤦

rosario (approved):

server/middleware/get-syndication-licence-for-user.js
```diff
@@ -54,6 +56,11 @@ module.exports = exports = async (req, res, next) => {
 
 		res.locals.licence = syndicationLicence;
 
+		if(!res.locals.hasGraphicSyndication) {
+			res.locals.hasGraphicSyndication = hasGraphicsAccess(syndicationLicence.products);
```
tolujimoh, line 60:
```suggestion
            const {GRAPHIC_SYNDICATION_PRODUCT_CODE} = require('config');
			res.locals.hasGraphicSyndication = syndicationLicence.products.some(product => product.code === GRAPHIC_SYNDICATION_PRODUCT_CODE);
```

This logic can be contained in the same file

server/middleware/get-contract-by-id-from-session.js
```diff
@@ -28,6 +29,10 @@ module.exports = exports = async (req, res, next) => {
 					acc[property] = acc[property] || assets.some(({ content }) => content.toLowerCase().includes(content_area));
 				});
 
+				if(hasGraphicSyndication) {
+					acc['rich_articles'] = hasGraphicSyndication;
+				}
+
 				return acc;
 			}, {
 				contributor_content: contract.contributor_content
```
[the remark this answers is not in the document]
thurstontye: Yeah that is much better

server/lib/get-contract-by-id.js
```diff
@@ -36,8 +37,13 @@ function decorateContract(contract) {
 	}
 
 	contract.itemsMap = contract.items.reduce((acc, asset) => {
+
 		if (asset.download_limit > 0) {
-			contentAllowed.push(ASSET_TYPE_TO_DISPLAY_TYPE[asset.asset_type]);
+			let contentTypeLabel = asset.asset_type;
+			if(hasGraphics && asset.asset_type === 'FT Article') {
+				contentTypeLabel  = 'FT Rich Article';
```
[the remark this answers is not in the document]
thurstontye: noted

config/default.yaml
```diff
@@ -10,6 +10,7 @@
     FT Article: Articles
     Podcast: Podcasts
     Video: Videos
+    FT Rich Article: Rich Articles
```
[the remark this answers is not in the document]
thurstontye: Fair comment thought I needed to but on reflection it's not neccessary

server/middleware/get-syndication-licence-for-user.js
```diff
@@ -54,6 +56,11 @@ module.exports = exports = async (req, res, next) => {
 
 		res.locals.licence = syndicationLicence;
 
+		if(!res.locals.hasGraphicSyndication) {
```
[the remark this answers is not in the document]
thurstontye: true

server/middleware/get-syndication-licence-for-user.js
thurstontye: True did this becuase I thought I was going to need to repeat. It's tested as a standalone helper so don't think theres any harm in leaving it so.

tolujimoh (approved):

PR #264, opened 2020-10-23
Add Logger class

rosario:
Added a Logger class similar to `next-profile` and `next-retention`. 

Replaced old logging with a log instance of Logger including `uuid` when possibile and function/component names

### DIFF
--- a/db/pg/index.js
+++ b/db/pg/index.js
@@ -2,7 +2,7 @@
 
 const path = require('path');
 
-const log = require('../../server/lib/logger');
+const { Logger } = require('../../server/lib/logger');
 
 const massive = require('massive');
 const pgConn = require('pg-connection-string');
@@ -10,6 +10,7 @@ const pgConn = require('pg-connection-string');
 const { DB } = require('config');
 
 const MODULE_ID = path.relative(process.cwd(), module.id) || require(path.resolve('./package.json')).name;
+const log = new Logger({source: MODULE_ID});
 
 let db;
 
--- a/server/controllers/resolve.js
+++ b/server/controllers/resolve.js
@@ -16,7 +16,7 @@ module.exports = exports = async (req, res, next) => {
 	const { locals: { contract, flags } } = res;
 
 	if (!Array.isArray(body)) {
-		log.warn(`Expected \`req.body\` to be [object Array] and got \`${Object.prototype.toString.call(body)}\` instead`, {
+		log.warn('RESOLVE_INVALID_REQUEST_BODY', {
 			message: `Expected \`req.body\` to be [object Array] and got \`${Object.prototype.toString.call(body)}\` instead`,
 			referer: req.headers.referer
 		});
@@ -25,7 +25,7 @@ module.exports = exports = async (req, res, next) => {
 	}
 
 	if (!body.length) {
-		log.error('`req.body` does not contain any content IDs', {
+		log.error('RESOLVE_BODY_NO_CONTENT_ID', {
 			message: '`req.body` does not contain any content IDs',
 			referer: req.headers.referer
 		});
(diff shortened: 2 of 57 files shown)


db/pg/index.js
```diff
@@ -2,14 +2,15 @@
...
 
-const log = require('../../server/lib/logger');
+const { Logger } = require('../../server/lib/logger');
 
 const massive = require('massive');
 const pgConn = require('pg-connection-string');
 
 const { DB } = require('config');
 
 const MODULE_ID = path.relative(process.cwd(), module.id) || require(path.resolve('./package.json')).name;
```
tolujimoh, line 12:
```
const MODULE_ID = path.relative(process.cwd(), module.id) || require(path.resolve('./package.json')).name;
```
is repeated in all the files that call the logger. This can be moved into the logger class. The logger class would be instantiated like this, passing in the module.id.

```
const log = new Logger({sourceID: module.id});
```

server/controllers/resolve.js
```diff
@@ -9,13 +9,14 @@ const getAllExistingItemsForContract = require('../lib/get-all-existing-items-fo
...
 module.exports = exports = async (req, res, next) => {
+	const log = new Logger({req, res, source: 'controllers/resolve'});
 
 	const { body } = req;
 
 	const { locals: { contract, flags } } = res;
 
 	if (!Array.isArray(body)) {
-		log.warn({
+		log.warn(`Expected \`req.body\` to be [object Array] and got \`${Object.prototype.toString.call(body)}\` instead`, {
```
tolujimoh, line 19: Preferably, this should be a shorter phrase without spaces or punctuation (using underscores like `RESOLVE_INVALID_REQUEST_BODY` or camelCase).

The detailed message is already contained in the data object. Also, this would make writing queries and grouping errors together easier.

server/controllers/resolve.js
```diff
@@ -24,7 +25,7 @@ module.exports = exports = async (req, res, next) => {
 	}
 
 	if (!body.length) {
-		log.error({
+		log.error('`req.body` does not contain any content IDs', {
```
tolujimoh, line 28: Same as above

server/lib/reformat-salesforce-contract.js
```diff
@@ -32,7 +33,7 @@ module.exports = exports = SFContract => {
 			const asset = findMatchingAsset(addendum, contract);
 
 			if (!asset) {
-				log.error({
+				log.error('Asset not found for Addendum', {
```
tolujimoh, line 36: underscore or camelCase. We should decide on one and kept it consistent.

tolujimoh (commented):
Looks good 👍 
Added a few comments

server/lib/logger.js
```diff
@@ -1,10 +1,120 @@
...
+class Logger {
+
+	/**
+	 * @param {object} req Express request object
+	 * @param {object} res Express response object
+	 * @param {string} source Where the log will be called from.
+	 */
+	constructor ({req, res, source}) {
+		// this.isTest = isTest(req);
+		// this.metricEnvironment = this.isTest ? 'test' : 'prod';
```
tolujimoh, line 23: Redundant comments can remove.

db/pg/index.js
rosario: Unfortunately not, there are many files without the `MODULE_ID`.

For example all the files in `/controllers/*`

https://github.com/Financial-Times/next-syndication-api/pull/264/files#diff-e25960b81c24078fa56a96ecbdf30b70a9d69ea20b5bc99e8ba3d8c15ab8d507R12

tolujimoh (approved):
Ship it! 🚀

PR #275, opened 2020-11-23
ACC-754 fixed evil JS idiosyncrasy 

AniaMakes:
From MDN docs: 

> every acts like the "for all" quantifier in mathematics. In particular, for an empty array, it returns true. (It is vacuously true that all elements of the empty set satisfy any given condition.)

So instead `array && array.every` we need to have `array.length > 0 && array.every`

### DIFF
--- a/server/lib/enrich/article.js
+++ b/server/lib/enrich/article.js
@@ -39,8 +39,8 @@ module.exports = exports = function article(content, contract, graphicSyndicatio
 			)
 
 	content.canAllGraphicsBeSyndicated =
-		// From MDN docs: every acts like the "for all" quantifier in mathematics. 
-		// In particular, for an empty array, it returns true. 
+		// From MDN docs: every acts like the "for all" quantifier in mathematics.
+		// In particular, for an empty array, it returns true.
 		// So if the filter above returned an empty array, [].every(whatever) would return `true`
 		graphicEmbeds.length > 0 &&
 		graphicEmbeds
--- a/server/lib/enrich/article.js
+++ b/server/lib/enrich/article.js
@@ -39,7 +39,10 @@ module.exports = exports = function article(content, contract, graphicSyndicatio
 			)
 
 	content.canAllGraphicsBeSyndicated =
-		graphicEmbeds &&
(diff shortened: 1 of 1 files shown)


server/lib/enrich/article.js
```diff
@@ -58,7 +61,7 @@ module.exports = exports = function article(content, contract, graphicSyndicatio
 			.removeElementsByTagName()
 			.removeProprietaryElement();
 
-		if (graphicSyndicationFlag && content.extension  === 'docx' && contract.allowed.rich_articles){
+		if (graphicSyndicationFlag && content.extension === 'docx' && contract.allowed.rich_articles) {
```
AniaMakes, line 64: Linter fixed an extra space on save.

server/lib/enrich/article.js
```diff
@@ -39,7 +39,10 @@ module.exports = exports = function article(content, contract, graphicSyndicatio
 			)
 
 	content.canAllGraphicsBeSyndicated =
-		graphicEmbeds &&
+		// From MDN docs: every acts like the "for all" quantifier in mathematics.
+		// In particular, for an empty array, it returns true.
+		// So if the filter above returned an empty array, [].every(whatever) would return `true`
+		graphicEmbeds.length > 0 &&
```
rosario, line 45: When `graphicEmbeds` is undefined this will break. We need to keep the previous check

server/lib/enrich/article.js
```diff
@@ -39,7 +39,10 @@ module.exports = exports = function article(content, contract, graphicSyndicatio
...
 		graphicEmbeds
 			.every((item) => item.canBeSyndicated === 'yes');
```
tolujimoh, line 47: Good spot! 👍 

Should we first check that `graphicEmbeds` is not false as well i.e,

```js
graphicEmbeds &&
graphicEmbeds.length > 0 &&
graphicEmbeds.every((item) => item.canBeSyndicated === 'yes');
```

Given that `graphicEmbeds` could be returned as `false` if embeds array isn't present in content.

AniaMakes: Fair point!

tolujimoh (approved):
:shipit:

PR #279, opened 2020-12-22
Update runbook to clarify how to call the API

AniaMakes:
The previous version only had "GET the URL with API key" but no info where one finds the key or what format the contract ID needs to be in. Given this is intended for OpsCops / emergency use, I felt it was prudent to add those to be explicit.


thurstontye (approved):
👍  Thanks for updating this.

PR #281, opened 2021-01-12
ACC-798 remove erasure endpoint 

oluoluoxenfree:
Removing the erasure endpoint from next-syndication-api; it's no longer in use and is causing errors

> In Splunk (index=restricted_ftdotcom source="133g5bgac00hv4v8t0dmry" request.path="/syn*") there have only been 6 requests to the erasure endpoint in the last week so it looks like it’s something we migrated and forgot to tidy up.
> (It’s been broken since 11th Nov last year when this PR was merged and nobody has complained loudly about missing it)


binaryberry:
Would be worth getting @tolujimoh 's opinion on this one if he can remember why he'd added that package in the first place while upgrading n-internal-tools: https://github.com/Financial-Times/next-syndication-api/pull/222

binaryberry:
Ha! this might be your best best actually Olu: https://financialtimes.slack.com/archives/C041V9QA7/p1610523704154600?thread_ts=1610470791.150700&cid=C041V9QA7

tolujimoh (commented):
This files ([erasure view](https://github.com/Financial-Times/next-syndication-api/blob/37be56f056368ec45c124d246d63c5905a4ddc78/views/erasure.html), [erasure search results view](https://github.com/Financial-Times/next-syndication-api/blob/37be56f056368ec45c124d246d63c5905a4ddc78/views/erasure-search-results.html) and [erasure controller](https://github.com/Financial-Times/next-syndication-api/blob/remove-s3o/server/controllers/erasure-search.js) ) should also be removed as they are now redundant

oluoluoxenfree:
oops, good spot!

tolujimoh (approved):
👍

PR #284, opened 2021-02-09
ACC-798 Update massive implementation by removing use of `db.reload()`

oluoluoxenfree:
Ticket: https://financialtimes.atlassian.net/secure/RapidBoard.jspa?rapidView=1108&projectKey=ACC&modal=detail&selectedIssue=ACC-798

Description:

Removes calls to `db.reload`, a function of [massive.js](https://massivejs.org/). 

We shouldn't be calling reload as it says in the documentation that [it is only necessary when changing the schema](https://massivejs.org/docs/connecting#reloading-the-api), which we do not do. Calling reload in the `else` block of `pg/index` meant that the database was being reloaded hundreds of times an hour, when changes aren't necessarily being made.  

A healthcare check on this package is constantly firing as a result of [a collision error that fires when a function has already been assigned](https://gitlab.com/dmfay/massive-js/-/blob/master/lib/database.js#L158), and this will hopefully go some way towards solving it.

### DIFF
--- a/db/pg/index.js
+++ b/db/pg/index.js
@@ -41,11 +41,5 @@ module.exports = exports = async (options = DB) => {
 			db = await massive(conn);
 		}
 	}
-	else {
-		log.info(`${MODULE_ID} reloading DB instance => `, db.instance.$cn);
-
-		db = await db.reload();
-	}
-
 	return db;
 };


tolujimoh (commented):
I don't think we need to add a `db.reload()` after every update operation. [This bit](https://massivejs.org/docs/connecting#reloading-the-api) in massive's documentation says to use `db.reload()` after we update the database schema. 

These update operations do not change the database schema, as such, does not require the db to be reloaded afterwards

binaryberry (changes requested):
Nice one Olu! See Tolu's comment about removing the reload rather than moving it. 

Could you explain in the commit message and PR description why we want to remove the use of db.reload, namely because it's only necessary when changing the schema (link to massive docs) and that calling it in the pg/index file means we're reloading the db hundreds of times an hour when no changes have necessarily been made to it, and that's triggering some healthchecks due to errors from the massive package.

binaryberry (approved):
Looks good to me!

PR #292, opened 2021-03-22
Upgrade n-express 19.23.1 -> 21.0.0

andygout:
This PR upgrades `n-express` to the latest major version. 

It cannot yet be pushed to the most recent patch version (v20.0.9) because it must first upgrade its `n-es-client` version to v3.0.0 so that it is only making requests to the Elasticsearch v7 cluster.

The major upgrades that have occurred are:
- **v20.0.0: Upgrade to Node v12** - this app runs on Node v12 so this means it is consuming a package that has been tested on the same version
- **v21.0.0: Upgrade n-health ^3.0.0 -> ^5.0.0**
  - **`n-health` v4.0.0: Upgrade to Node v12** - same as above, i.e. no issue
  - **`n-health` v5.0.0: deprecate Pingdom healthcheck**

> this change removes the code that's actually checking Pingdom, and makes the Pingdom healthcheck class always fail, so that apps that inadvertently update to this version without migrating the check don't crash on startup, but instead get an alert that they can easily action.

> this check will be fully removed in the next major version of n-health, which will be released when enough apps have updated to this version and migrated their checks.

> #### migration

> when updating to this release, if you're using any Pingdom healthchecks, they'll start to unconditionally fail. you'll need to:

> 1. remove these checks from your healthcheck configuration
> 2. update the checks themselves in Pingdom to tag them with your system code (see step 8 in the "How To Pingdom Check" guide)

> note that if you're using healthchecks to monitor other teams' systems, you'll inherit the severity they've set for their own Pingdom checks. we recommend duplicating this check for your own visibility, so you can set the severity and tagging for yourself, or writing a custom healthcheck to make calls similar to those your app is making.

---

> 1. remove these checks from your healthcheck configuration

Removed in this PR.

> 2. update the checks themselves in Pingdom to tag them with your system code (see step 8 in the "How To Pingdom Check" guide)

As follows:

---

#### Pingdom check for [User Rights US Service reachable (check 7834166)](https://my.pingdom.com/reports/responsetime#daterange=7days&tab=uptime_tab&check=7834166)
Previously used [check 4897678](https://my.pingdom.com/reports/responsetime#daterange=7days&tab=uptime_tab&check=4897678)
4897678-to-7834166

---

#### Pingdom check for [User Rights EU Service reachable (check 7834226)](https://my.pingdom.com/reports/responsetime#daterange=7days&tab=uptime_tab&check=7834226)
[...]

PR #293, opened 2021-03-22
Upgrade n-es-client 1.8.0 -> 3.0.0; n-express 21.0.0 -> 21.0.9

andygout:
**This PR is dependent on https://github.com/Financial-Times/next-syndication-api/pull/292, which should be merged into the main branch first to implement the required changes for the `n-health` major upgrade, followed by a rebase of this branch onto the main branch (to pick up those changes).**

This PR upgrades the `n-es-client` dependency to v3.0.0 so that all of its Elasticsearch requests are directed to the new Elasticsearch v7 cluster.

It also upgrades the `n-express` version (which in turn upgrades the `next-metrics` version) so that metrics for requests to the Elasticsearch v7 cluster are captured correctly.

This app only uses `n-es-client` to make `get` and `mget` requests and the implementation applied here for those methods does not change between versions, meaning no further changes are required.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,8 +4,8 @@
   "version": "0.35.1",
   "private": true,
   "dependencies": {
-    "@financial-times/n-es-client": "1.8.0",
-    "@financial-times/n-express": "21.0.0",
+    "@financial-times/n-es-client": "3.0.0",
+    "@financial-times/n-express": "21.0.9",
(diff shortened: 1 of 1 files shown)


AniaMakes (approved):
I'm happy with those changes - we ought to babysit the graphs once this is merged, just in case.

PR #297, opened 2021-03-25
NOPS-602 Enabled debugging for mail-contributor

AniaMakes:
[Ticket](https://financialtimes.atlassian.net/browse/NOPS-602)

**TL;DR of the issue:** When an article by a freelancer gets syndicated, and email gets sent out to notify Someone that they need to pay extra money to the freelancer. These emails have stopped some time ago and need to be reinstated. 

**What we've done:**
- We are unable to see any of the logs in the `mail-contributor` file. We suspect this may be because the `logs` have been disabled in the config file. This PR enables the logs to see if that helps. 
- We are unsure if the function as a whole is getting called at all, and if yes, does it `return` at a first possible step (ie - none of the articles syndicated are written by freelancers). This PR adds `log.info` immediately after initiation to allow us to see if this function is getting called, and with what `event`.

### DIFF
--- a/config/production.yaml
+++ b/config/production.yaml
@@ -3,7 +3,7 @@
   CONTRIBUTOR_EMAIL:
     transport:
       debug: false
-      logger: false
+      logger: true
 
   CONVERT_FORMAT_COMMAND: "/app/pandoc-dpkg/usr/bin/pandoc"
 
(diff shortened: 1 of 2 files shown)


tolujimoh (approved):

config/production.yaml
```diff
@@ -3,7 +3,7 @@
   CONTRIBUTOR_EMAIL:
     transport:
       debug: false
-      logger: false
```
binaryberry, line 6: Is that going to log a contributor's personal email address to Splunk?

binaryberry: (I think it would make GDPR sad)

AniaMakes: I doubt it, it looks to be the name of the variable https://github.com/Financial-Times/next-syndication-api/blob/master/worker/sync/db-persist/mail-contributor.js#L14

PR #301, opened 2021-04-19
NOPS-602 email client makeover for mail-contributor

AniaMakes:
[Ticket](https://financialtimes.atlassian.net/browse/NOPS-602)

Issue: Due to gmail security updates, gmail blocked `nodemailer` attempts to log in to sock-puppet send an email. As such, relevant alerts to pay the freelancers didn't get sent. 

This PR: 
- removes `nodemailer` functionality from `mail-contributor.js` and replaces it with [e-mail platform's POST send by address](https://ep.ft.com/docs/email-send-api/#api-Send-Post_send-by-address)
- you'll notice there's the ACCOUNTS_EMAIL as one of the recipients. I set that to my email address so that I get to see a copy of the email. Once I'm happy it looks good, I'll remove it. 

This PR does NOT:
- remove modules and google environment variables. My linter does not work for this repo and I am not confident to do this without a machine shouting at me if I do something wrong.  
- check if nodemailer is used elsewhere / replace that. Ticket is only to deal with withContributor emails getting missed, so anything else should be a separate ticket.

### DIFF
--- a/test/worker/sync/db-persist/mail-contributor.spec.js
+++ b/test/worker/sync/db-persist/mail-contributor.spec.js
@@ -21,7 +21,7 @@ chai.use(sinonChai);
 
 const MODULE_ID = path.relative(`${process.cwd()}/test`, module.id) || require(path.resolve('./package.json')).name;
 
-describe.only(MODULE_ID, function () {
+describe(MODULE_ID, function () {
 	const contractResponse = require(path.resolve(`${FIXTURES_DIRECTORY}/contractResponse.json`));
 	let underTest;
 	let db;
--- a/test/worker/sync/db-persist/mail-contributor.spec.js
+++ b/test/worker/sync/db-persist/mail-contributor.spec.js
@@ -6,6 +6,7 @@ const chai = require('chai');
 const sinon = require('sinon');
 const sinonChai = require('sinon-chai');
 const proxyquire = require('proxyquire');
+const fetchMock = require('fetch-mock').sandbox();
 
 const NodeMailerJSONTransport = require('nodemailer/lib/json-transport');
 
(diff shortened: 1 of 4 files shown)


test/worker/sync/db-persist/mail-contributor.spec.js
```diff
@@ -21,7 +21,7 @@ chai.use(sinonChai);
 
 const MODULE_ID = path.relative(`${process.cwd()}/test`, module.id) || require(path.resolve('./package.json')).name;
 
-describe(MODULE_ID, function () {
+describe.only(MODULE_ID, function () {
```
joelcarr, line 24: Left over `.only`

joelcarr (commented):
Looking good as far as I can understand.

> you'll notice there's the ACCOUNTS_EMAIL as one of the recipients. I set that to my email address so that I get to see a copy of the email. Once I'm happy it looks good, I'll remove it.

Do we have an accounts team email this could be instead? Just incase someone else has to pick up the work all of a sudden.

test/worker/sync/db-persist/mail-contributor.spec.js
AniaMakes: Ugh, that's why I hate linter not working. Didn't even catch this on push.

thurstontye:
> Looking good as far as I can understand.
> 
> > you'll notice there's the ACCOUNTS_EMAIL as one of the recipients. I set that to my email address so that I get to see a copy of the email. Once I'm happy it looks good, I'll remove it.
> 
> Do we have an accounts team email this could be instead? Just incase someone else has to pick up the work all of a sudden.

We have ft-accounts-team@ft.com  but do we want these emails going to a group account?

joelcarr (approved):
✉️ 💅

PR #302, opened 2021-04-20
ACC-1005 Added a PR template

AniaMakes:
Following https://github.com/Financial-Times/incident-reports/issues/441 it became obvious that the special release / deploy procedure isn't documented visibly enough. To try to prevent similar issues in the future, this PR introduces a template that anyone raising the PR should follow - this should highlight the special procedure.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,11 @@
+### Description
+<!--- Describe your changes in detail -->
+
+### Ticket
+<!-- Add link to the corresponding ticket -->
+
+### What is the new version number in package.js?
+<!-- If you have made any changes other than documentation, you need to follow a special deploy process, as described here https://github.com/Financial-Times/next-syndication-dl#deploying-the-download-server -->
+
+### Link to the next-syndication-dl PR which bumps the next-syndication-api version
+<!-- Add link to the PR -->


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,11 @@
+### Description
+<!--- Describe your changes in detail -->
+
+### Ticket
+<!-- Add link to the corresponding ticket -->
+
+### What is the new version number in package.js?
+<!-- If you have made any changes other than documentation, you need to follow a special deploy process, as described here https://github.com/Financial-Times/next-syndication-dl#deploying-the-download-server -->
```
valeksiev, line 8: Obviously, I am not a native speaker but how about something amongst the lines below?
```suggestion
<!-- If you have made any changes other than documentation, please consider going through the deployment process described here: https://github.com/Financial-Times/next-syndication-dl#deploying-the-download-server. Any changes made to functionality used explicitly or implicitly by `next-syndication-api/server/app-dl.js` and `server/controllers/download-by-content-id.js` will most probably need to make their way to Heroku's deployment of `next-syndication-dl` as soon as they do for `next-syndication-api` -->

```

AniaMakes: I considered being a bit more specific with what changes need to be officially released, but I think it's better to err on the side of caution - saying changes to any functionality will lead to always bumping the version and not being blocked by "does this count" - syndication is still a bit of black box, so it's safer to always make a release.

valeksiev (approved):

PR #305, opened 2021-05-12
NOPS-602 Email makeover take 2: log and env var fixes

AniaMakes:
### Description
Follow up to #301 

Today we had the first instance when an email should have been sent - but sadly wasn't. Our logs were broken because `res.json()` returns a Promise but we weren't handling that so it was empty. However Andreas from Email Platforms was able to report that the call was made and that the error was "Cannot read property 'email' of null".

This PR:
- moves `res.json()` out of a log and handles it properly
- adds a log for the body and headers we're sending so we can see exactly what we're doing
- adds ACCOUNTS_EMAIL to the custom environment variables yaml

### Ticket
https://financialtimes.atlassian.net/browse/NOPS-602

0.38.0

https://github.com/Financial-Times/next-syndication-dl/pull/57 (CI fails because we haven't merged this yet, so it can't build)


magsallen (approved):
This looks good. It's a very neat PR 🎉

PR #306, opened 2021-05-13
NOPS-602 race condition fix

AniaMakes:
### Description
Follow up to #305 

I merged before Tolu had a chance to comment and he pointed out that my `res.json()` fix will hit a race condition and suggested simplifying to `await res.json()`

This PR fixed the dodgy fix to `res.json()`.

### Ticket
https://financialtimes.atlassian.net/browse/NOPS-602

0.39.0 (because I already released the 0.38.0

I'm going to change this one to 0.39.0
https://github.com/Financial-Times/next-syndication-dl/pull/57

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "ft-next-syndication-api",
   "description": "Next Syndication API",
-  "version": "0.39.0",
+  "version": "0.38.1",
   "private": true,
   "dependencies": {
     "@financial-times/n-es-client": "3.0.0",
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "ft-next-syndication-api",
   "description": "Next Syndication API",
-  "version": "0.38.0",
+  "version": "0.39.0",
   "private": true,
   "dependencies": {
     "@financial-times/n-es-client": "3.0.0",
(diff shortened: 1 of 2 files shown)


worker/sync/db-persist/mail-contributor.js
```diff
@@ -74,8 +74,7 @@ module.exports = exports = async (event) => {
 			log.info('Some emails were not delivered. Not delivered count', res.total_rejected_recipients)
 		}
 
-		let responseData;
-		res.json().then(data => responseData = data)
+		let responseData = await res.json();
```
tolujimoh, line 77: improvement: Should be `const` as the variable is not reassigned
```suggestion
		const responseData = await res.json();
```

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "ft-next-syndication-api",
   "description": "Next Syndication API",
-  "version": "0.38.0",
+  "version": "0.39.0",
```
tolujimoh, line 4: nitpick: Should be a patch release as it is a bug fix
https://semver.org/#spec-item-6

```suggestion
  "version": "0.38.1",
```

tolujimoh (approved):
💯

PR #309, opened 2021-07-05
ACC-1106 SSL fine tune for DB connection

AniaMakes:
### Description
**Problem at hand:** we are seeing A LOT of `no pg_hba.conf` errors in Splunk. 

**Potential culprit:** SSL needs to be turned on [according to Heroku docs](https://help.heroku.com/DR0TTWWD/seeing-fatal-no-pg_hba-conf-entry-errors-in-postgres)

**But...** ... we already have `rejectUnauthorized: false` in place, per instructions. 

**Thought process**:
Remote connections in heroku need an extra parameter in the URL [doc link](https://devcenter.heroku.com/articles/heroku-postgresql#remote-connections). This would be the simple solution, as explained in this [forum thread](https://www.reddit.com/r/Heroku/comments/f3xm3w/fatal_no_pg_hbaconf_entry_for_host/) if it wasn't for the fact that we are using a `pg-promise configuration object` rather that the `connection string`.

Spiderman meme time: Heroku explicitly states "it is important to add this parameter in code rather than editing the config var directly" [doc link](https://devcenter.heroku.com/articles/heroku-postgresql#remote-connections). At the same time they say "Alternatively, you can omit the ssl configuration object if you specify the `PGSSLMODE` config var: `heroku config:set PGSSLMODE=no-verify`" [doc link, in the blue megaphone section](https://devcenter.heroku.com/articles/heroku-postgresql#connecting-in-node-js)

 I checked if others had the same problem and the suggested solution is one I ended up using as described in this [github issue thread](https://github.com/nextauthjs/next-auth/issues/822#issuecomment-723171719) - there's a bunch of people who say this worked for them. 

**Why I think this will work**: 
The `pg-promise configuration object` accepts a bunch of parameters. The [docs](https://github.com/vitaly-t/pg-promise/wiki/Connection-Syntax#configuration-object) say that "there are more advances parameters" and links to the repo of the [driver module](https://github.com/brianc/node-postgres). Now, looking in the [pg-connection-string repo](https://github.com/brianc/node-postgres) - I KNOW WE ARE NOT USING STRING - it says "any other query parameters are preserved intact". In face of this, I am going to hazard a guess that whichever of the repos deals with the `pg-promise config objects` uses the same "preserve intact" logic. 

**What if it doesn't work?**
We roll back 🤷🏻‍♂️  I can't think of anything else that would work. 

**Anything else?**
I used the spread operator, so I bumped up the ecma version in eslint. 
I also got fed up of linter running on pre-push and forcing commits with typos or playing un-commit game, so I added it to run on pre-commit. 

### Ticket
https://financialtimes.atlassian.net/browse/ACC-1106

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "ft-next-syndication-api",
   "description": "Next Syndication API",
-  "version": "0.38.1",
+  "version": "0.38.2",
   "private": true,
   "dependencies": {
     "@financial-times/n-es-client": "3.0.0",


db/pg/index.js
```diff
@@ -12,15 +12,23 @@ const { DB } = require('config');
...
 let db;
 
 module.exports = exports = async (options = DB) => {
 	if (!db || options !== DB) {
 		log.info(`${MODULE_ID} creating new DB instance with options => `, options);
 
 		if (options.uri) {
-			const conn = Object.assign({ ssl: { rejectUnauthorized : false } }, pgConn.parse(options.uri));
-
+			const conn = Object.assign({ ...sslSetUp}, pgConn.parse(options.uri));
```
joelcarr, line 31: **Question**: Are we expecting the addition to the target object of the following?

from:

```js
{ 
    ssl: { 
        rejectUnauthorized : false 
    }  
}
```

to: 

```js
{
    ssl: true,
	extra: {
		ssl: {
			rejectUnauthorized: false,
		},
	}
} 
```

db/pg/index.js
```diff
@@ -31,13 +39,10 @@ module.exports = exports = async (options = DB) => {
 				host: options.host,
 				password: options.password,
 				port: options.port,
-				user: options.user_name
+				user: options.user_name,
+				...sslSetUp,
```
joelcarr, line 43: **Question**: Are we expecting the addition to the target object of the following?

from:

```js
{
    ..., 
    ssl: { 
        rejectUnauthorized : false 
    }  
}
```

to: 

```js
{
    ..., 
    ssl: true,
	extra: {
		ssl: {
			rejectUnauthorized: false,
		},
	}
} 
```

joelcarr: According to https://github.com/brianc/node-postgres/tree/master/packages/pg-connection-string#usage, the connection object does not expect a `extra: {...}`. I am struggling to find where `extra` could be used.

PR #310, opened 2021-07-07
Revert "ACC-1106 SSL fine tune for DB connection"

AniaMakes:
Reverts Financial-Times/next-syndication-api#309

The release has taken Syndication down with a massive spike of `error_message: self signed certificate`. Generally the recommended solution is to use `ssl = {rejectUnauthorised: false}` which is what we had to start with. 

I guess I don't need to worry about making a new release because `next-syndication-dl` `main` still looks for 0.38.1 and this PR reverts to that version?

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -8,7 +8,7 @@ const config = {
 		'node': true
 	},
 	'parserOptions': {
-		'ecmaVersion': 2018,
+		'ecmaVersion': 2017,
 		'sourceType': 'module'
 	},
 	'rules': {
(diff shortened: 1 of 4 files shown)


binaryberry:
Have we ruled out the version number having an extra dot or syndication-dl being out of sync thing as potential causes? Or is this a temporary thing to put Heroku and github in sync until we figure that out?

joelcarr (commented):
We make this a new release `v0.38.3` or delete `v0.38.2`. I feel like creating a new release is the most "proper" way to go (acting as if this was a widely used open source library). What are your opinions?

AniaMakes:
@binaryberry I can make another release without the extra dot, but we probably should tell Jenn and keep finger ready on rollback button? It feels Syndication should not break this spectacularly if the repos are out of sync (they generally don't).

binaryberry (approved):

joelcarr (approved):
⏪

PR #312, opened 2021-07-08
ACC-1106 SSL fix, take two

AniaMakes:
### Description
#309 took Syndication down, so it's probably not the way to go. See that PR for the background for this ticket. 

This PR takes a wild guess that `massive.js` dependency will pick out `sslmode` from the `ssl` object based on this piece of code https://github.com/brianc/node-postgres/blob/master/packages/pg/lib/connection-parameters.js#L137

This PR add `sslmode: require` to `ssl` object. Heroku [docs](https://devcenter.heroku.com/articles/heroku-postgresql#remote-connections) state `sslmode: require` is needed for remote connections. 

### Ticket
https://financialtimes.atlassian.net/browse/ACC-1106

0.38.4

Will wait to make sure this PR doesn't break Syndication before making a PR in DL.

### DIFF
--- a/db/pg/index.js
+++ b/db/pg/index.js
@@ -19,7 +19,7 @@ module.exports = exports = async (options = DB) => {
 		log.info(`${MODULE_ID} creating new DB instance with options => `, options);
 
 		if (options.uri) {
-			const conn = Object.assign({ ssl: { rejectUnauthorized : false } }, pgConn.parse(options.uri));
+			const conn = Object.assign({ ssl: { rejectUnauthorized : false, sslmode: require } }, pgConn.parse(options.uri));
 
 			log.info(`${MODULE_ID} creating new DB instance with URI String => `, conn);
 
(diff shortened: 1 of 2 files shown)


db/pg/index.js
```diff
@@ -19,7 +19,7 @@ module.exports = exports = async (options = DB) => {
 		log.info(`${MODULE_ID} creating new DB instance with options => `, options);
 
 		if (options.uri) {
-			const conn = Object.assign({ ssl: { rejectUnauthorized : false } }, pgConn.parse(options.uri));
+			const conn = Object.assign({ ssl: { rejectUnauthorized : false, sslmode: require } }, pgConn.parse(options.uri));
```
joelcarr, line 22: **Issue:** `require` is a function (commonjs' [`require('module')`](https://nodejs.org/docs/latest/api/modules.html#modules_require_id)) here. This should be set as a string.

AniaMakes: Ooof, good catch 🤦🏻

AniaMakes:
This got fixed in Heroku, see #317

PR #316, opened 2021-07-20
Housekeeping on logs and healthchecks

AniaMakes:
### Description
This PR
- adds a NODE_ENV info to our logs to help with [ACC-1106](https://financialtimes.atlassian.net/browse/ACC-1106). Dawn spotted that the bucket names are different between the calls that are failing vs those that are successful which imply that maybe the way our secrets are loaded may be related? 
- adds IDs to health checks because Dawn asked why things are green, and well, without the ID the connection to the data source is questionable (the graph doesn't work, so I'm guessing the values may not be correct either) and the checks aren't complaint anyway. See [FT Health Check standard document](https://docs.google.com/document/d/18hefJjImF5IFp9WvPAm9Iq5_GmWzI9ahlKSzShpQl1s/edit) for more info on health checks. 

Screenshot 2021-07-20 at 11 31 03

### Ticket
n/a

This is logs and housekeeping, release isn't really needed.

### DIFF
--- a/db/pg/index.js
+++ b/db/pg/index.js
@@ -7,7 +7,7 @@ const { Logger } = require('../../server/lib/logger');
 const massive = require('massive');
 const pgConn = require('pg-connection-string');
 
-const { DB } = require('config');
+const { DB, NODE_ENV } = require('config');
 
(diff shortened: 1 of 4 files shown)


binaryberry (approved):
Nice!

joelcarr (approved):
㏒㏒㏒ 🤒

PR #319, opened 2021-08-12
Add a stop-local command to stop all clusters when running locally

debugwand:
### Description
We had a command to run locally which set up a bunch of clusters, but no corresponding stop command to shut down the clusters. This meant that even when you exited the app with `Ctrl-C`, there would still be clusters running the app. This made a clean restart difficult. 

### Ticket
None, minor fix

This is not something to be used in production, its a command for operating the app locally

N/A

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,6 +38,9 @@ run-local:
 	./node_modules/.bin/pm2 start procfile.json
 	./node_modules/.bin/pm2 logs
 
+stop-local:
+	./node_modules/.bin/pm2 stop procfile.json
+
 run-monit:
 	./node_modules/.bin/pm2 monit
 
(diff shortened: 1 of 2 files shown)


AniaMakes (commented):
[Questions]

1️⃣  _ I can find some docs about `start` for `procfile`, but I'm struggling to find anything talking about `stop`. It's a reasonable assumption that we `stop` would do just that, I'm wondering if you know of any official docs.

2️⃣  _ we already have `make kill-all` - what would be the difference between `kill-all` and `stop-local`? I know that `kill` is a unix(?) command to terminate a process, but that's the extent of my knowledge about `kill`.

debugwand:
https://pm2.keymetrics.io/docs/usage/quick-start/

I missed the kill-all command in the makefile, will use that instead

PR #321, opened 2021-09-24
Db persist worker triggers

debugwand:
### Description
Add an endpoint where we can trigger the db-persist workers so we can test they write to the database for upgrade work

### Ticket
n/a, part of syndication database upgrade work

n/a this is for testing code, won't run in production

n/a this is for testing code, won't run in production

### DIFF
--- a/server/controllers/nonproduction/db-persist.js
+++ b/server/controllers/nonproduction/db-persist.js
@@ -1,0 +1,72 @@
+'use strict';
+
+const { Logger } = require('../../lib/logger');
+
+const ACL = {
+	user: false,
+	superuser: false,
+	superdooperuser: true
+};
+
+/* values for the post request
+* saved and deleted run in upsert history, started runs in mail contributor
+* you can also override any of the base dummy event items
+* expected body:
+{
+	"state": "[saved|deleted|started]",
+	"user": {
+		"first_name": stringValue,
+		"surname": stringValue,
+		"email": stringValue,
+		"id": valid User ID that is already in the syndication database (e.g. any dev with S1 product)
+	},
+	"_id": random ID string,
+	"content_id": content ID for an article,
+	"content_url": content url matching that article,
+	"user_id": same user ID as above
+}
+*/
+module.exports = exports = async (req, res) => {
+	const log = new Logger({req, res, source: 'controllers/nonproduction/db-persist'});
+	try {
+		const { locals: { user } } = res;
+
+		if (!user || ACL[user.role] !== true) {
+			res.sendStatus(401);
+
+			return;
+		}
+
+		const eventSettings = req.body;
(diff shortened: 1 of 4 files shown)


server/controllers/nonproduction/db-persist.js
```diff
@@ -0,0 +1,72 @@
...
+* you can also override any of the base dummy event items
+* expected body:
+{
+	"state": "[saved|deleted|started]",
+	"user": {
+		"first_name": stringValue,
+		"surname": stringValue,
+		"email": stringValue,
+		"id": valid User ID that is already in the syndication database (e.g. any dev with S1 product)
+	},
```
magsallen, line 22: [Suggestion] Add `role` to the expected user props. By default I believe we have 'user' level permissions, so we'll need to overwrite this locally when running the command. See the `ACL[user.role]` check on line 34.

magsallen (commented):
Are we planning to keep this file after the db upgraded work or will be need to delete it once the testing is complete? We probably don't want the hardcoded `baseDummyEvent` data with `contact_id` and `licence_id` to hang around without explanation

server/controllers/nonproduction/db-persist.js
debugwand: I've had a check and the res body values don't get written to locals. I solved this by updating my own role on the database to the `superdooperuser` setting, previously I'd just commented out this check and made sure not to commit it.

debugwand:
> Are we planning to keep this file after the db upgraded work or will be need to delete it once the testing is complete? We probably don't want the hardcoded `baseDummyEvent` data with `contact_id` and `licence_id` to hang around without explanation

it will be a useful hook for testing these workers, which currently doesn't exist.

its using the FT Staff license so we can't accidentally add anything to a real customer's data. We could put that into config if we think it will be something that might change, afaik these values don't change (unless someone accidentally deleted the account and we had to make a new one).

server/controllers/nonproduction/db-persist.js
magsallen: Makes sense 👍

magsallen:
Sorry! I've just remembered this PR. I'm still a bit concerned about committing the staff licence details, even though they're for testing and can't be used to affect users. Could we move them to vault maybe?

debugwand:
> Sorry! I've just remembered this PR. I'm still a bit concerned about committing the staff licence details, even though they're for testing and can't be used to affect users. Could we move them to vault maybe?

yeah I think vault is probably right

magsallen (approved):
I'm a little concerned about committing the staff licence details

debugwand:
> I'm a little concerned about committing the staff licence details

yep I used rebase to squish them out of existence in the git commit history

PR #335, opened 2021-12-21
[ACC-1242] Update to npm@7

aasanchez:
Some Changes:

- Migrate to npm 7
- Pin npm@7.20.2 and node@12.22.5
- Remove update npm jobs to avoid problems with [old-openssh-version-on-circleci](https://github.com/financial-times/next/wiki/npm-7-Migration-Guide#old-openssh-version-on-circleci)
- Adding to git package-lock.json, for the same reason above.
- Update cache key to match with the npm version


polkwc (approved):
Well done!!

.circleci/config.yml
```diff
@@ -62,12 +65,11 @@ jobs:
...
+          version: "7"
       - run:
           name: Install project dependencies
           command: make install
-      - run:
-          name: shared-helper / npm-install-peer-deps
-          command: .circleci/shared-helpers/helper-npm-install-peer-deps
       - run:
           name: shared-helper / npm-update
           command: .circleci/shared-helpers/helper-npm-update
```
polkwc, line 75: **Question**: What about this update? according to we talked, I think this step has to be deleted

polkwc (commented):
I left a new comment. Could you check it?

.circleci/config.yml
aasanchez: Just fixed

PR #336, opened 2022-01-06
ACC-698 Add API healthchecks for 5xx error responses

asugar13:
### Description
Add API healthchecks for 5xx error responses.
Aiming to trigger when the 5xx/total response ratio is over 5%.
Will need to deploy in order to test.

This is the panel that contains the path that looks for 5xx error responses: https://grafana.ft.com/d/Px4RNC0nk/asier-playground?orgId=1&from=now-7d&to=now

### Ticket
[JIRA](https://financialtimes.atlassian.net/jira/software/c/projects/ACC/boards/1108?modal=detail&selectedIssue=ACC-698)

### DIFF
--- a/health/server-error-spikes.js
+++ b/health/server-error-spikes.js
@@ -1,0 +1,16 @@
+const nHealth = require('n-health');
+const threshold =5;
+const samplePeriod = 60; // minutes
+
+module.exports = nHealth.runCheck({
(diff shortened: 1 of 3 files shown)


GlynnPhillips:
Closing as this is very old and relies on Graphite which we are phasing out

PR #360, opened 2022-04-01
Add id property to database healthcheck

alexmuller:
### Description
Add an ID property to the database health check, to give us check history in Heimdall.

N/A - happy to let this get deployed with the next release

N/A - happy to let this get deployed with the next release


rowanmanning (approved):
🚀

PR #366, opened 2022-05-20
Remove two keys that are no longer in use

rowanbeentje:
### Description

As part of reviewing keys after the Heroku incident, I'm looking at some shared keys previously used for accessing the myFT API and splitting them out to unique per-service keys.

These two keys came up during my search, but in looking through the codebase they are no longer used - confirmed with a search for the exact key name and also for related myFT or prefs functionality.

### Ticket
(no ticket created!)

I looked at the deploy process and package bump guidance, but as this is only a minor cleanup with no behavioural changes, I believe this doesn't need to be deployed - it can just be included as part of the next release. So I have *not* altered the package.json version number and will not trigger a deploy post-merge.

n/a (see version number notes)


rowanbeentje:
(These keys are not in next-syndication-api vault - they're pulled in as part of the shared keys currently)

AniaMakes (approved):
Looks good, there will be more keys removed at a later point, but it's good to have a smaller batch.

PR #367, opened 2022-05-25
ACC-1518: standfirst missing on downloaded article

danieleruiz:
### Description
the sub headline (standfirst) is not being included when downloading an article
### Ticket
[TICKET](https://financialtimes.atlassian.net/browse/ACC-1518
)### What is the new version number in package.js?
0.41.2


server/lib/builders/document-builder.js
```diff
@@ -25,7 +25,7 @@ module.exports = exports = class DocumentBuilder {
 		this.content = content;
 
 		this.contentDocument = new DOMParser().parseFromString(
-			`<body>${content.bodyHTML}</body>`,
+			`<body><h3>${content.standfirst}</h3>${content.bodyHTML}</body>`,
```
jamesr101, line 28: question: is it possible an article won't have a `standfirst`?

danieleruiz: i don´t know but in that case it going to be parsed to a pdf without contend on the "h3". so i supposed will work anyway and looks ok

jamesr101 (approved):

server/lib/builders/document-builder.js
```diff
@@ -23,9 +23,9 @@ const RE_REMOVE_TAGS = /<\/?[^>]*>/gm;
 module.exports = exports = class DocumentBuilder {
 	constructor(content) {
 		this.content = content;
-
+		const standfirst = `<h3>${content.standfirst}</h3>`
```
asugar13, line 26: happy for this to be merged but it would be more readable like
const standfirst = content.standfirst ? `<h3>${content.standfirst}</h3>` : null
and then
`<body>${standfirst}${content.bodyHTML}</body>`

asugar13 (approved):

PR #372, opened 2022-06-02
Prevent app crashes caused by unhandled rejections 2

JSRedondo:
Description
These middlewares have always resulted in unhandled promise rejections.
Express (v4) can't handle thrown errors in asynchronous routes and so
these errors weren't handled correctly.

The behaviour for unhandled rejections changed in Node.js 16, they used
to just log a warning but now they result in app crashes. This is why we
haven't had lots of app crashes until recently. See the migration
notes

I'm now handling all the promise rejections properly by passing the
error onto the next route rather than rethrowing it. This should protect
the app from going down if there's an error in the routes.

Ticket
No ticket but it relates to [incident 1554](https://financialtimes.slack.com/archives/C03HF9PG7A9).


danieleruiz (approved):
thanks for that!

PR #373, opened 2022-06-13
FTDCS-152 - Remove body-parser dependency

JSRedondo:
Update the version n-express with text, json & urlencode exported functions.
Remove outdated package body-parser


AniaMakes (approved):
Looks reasonable. Thank you! 🧹

PR #392, opened 2022-09-05
ACC-1902 Bump NPM versions to support heroku logs drain

sakibshaik:
### Description
Heroku log drains are a more reliable way of getting our logs from our Node.js applications into Splunk. They are the preferred method supported by the Edge Delivery and Observability team ([docs](https://tech.in.ft.com/tech-topics/logging/splunk/logging-from-heroku)). They are better than making n-logger connect directly to Splunk because:

Logs sent by the application over HTTP may be batched and flushed, so in the event of an application crash we may not receive the final logs (to help diagnose the crash)

Log drains don’t use Node.js application resources (CPU and memory) to send logs

Log drains include Heroku API and system activity like application scaling, environment variable changes, and process exit codes

This PR is the first step of 3 that will allow the dot-com support team to migrate the app to use heroku drain logs in order to capture errors on why the app is crashing. This tends to use the Heroku API gateway to send logs to splunk as against using n-logger to send logs to splunk which makes the crashing error available on splunk logs.

[Read more here](https://financialtimes.atlassian.net/wiki/spaces/DS/pages/7883555001/Migrating+an+app+to+Heroku+log+drains)

### Ticket
https://financialtimes.atlassian.net/browse/ACC-1902


Nazehs (changes requested):
[Blocking] I think you forgot to update some of the packages stated on the [migration guide](https://financialtimes.atlassian.net/wiki/spaces/DS/pages/7883555001/Migrating+an+app+to+Heroku+log+drains) example packages are  n-health, n-eager-fetch, etc

Nazehs (approved):
Happy to have it merge

PR #398, opened 2022-10-04
chore(): bump n-gage to v9.x

jamesr101:
## Updating n-gage to latest Version v9.0.1

This to work on [this ticket ](https://financialtimes.atlassian.net/browse/ACC-1943?atlOrigin=eyJpIjoiN2U5ZDIxZmExOWVmNDg2Y2JiMGY2Mzg2YWNhMDA4MTgiLCJwIjoiaiJ9). In order to remove `NPM_CONFIG_PRODUCTION` from vault which was required, but fixed in n-gage v9. 

### Breaking changes for v9.0.1:
no longer supporting older versions of node and npm (this project does not use them)

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
     "xmldom": "^0.6.0"
   },
   "devDependencies": {
-    "@financial-times/n-gage": "^8.2.0",
+    "@financial-times/n-gage": "9.0.1",
     "@financial-times/n-heroku-tools": "^10.0.0",
     "@financial-times/n-test": "^1.13.2",
     "chai": "^4.0.2",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -33,7 +33,7 @@
     "xmldom": "^0.6.0"
   },
   "devDependencies": {
-    "@financial-times/n-gage": "^8.2.0",
```
binaryberry, line 36: [question, non blocking] is there a reason you're doing away with the caret? I feel like FT brewed packages can be trusted enough to follow semantic versioning

binaryberry (approved):
One question, but approved!

PR #418, opened 2022-12-13
ACC-2118 Capitalize first letter of time period before save to DB

asugar13:
### Description
If the contract data is coming all lowercase from Salesforce, capitalize the first letter before saving to the DB. This is because of the enum_time_period in next-syndication-db-schema, which only accepts uppercase: https://github.com/Financial-Times/next-syndication-db-schema/blob/main/schema.syndication.base/type/enum_time_period.sql

### Ticket
[JIRA](https://financialtimes.atlassian.net/jira/software/c/projects/ACC/boards/1461?modal=detail&selectedIssue=ACC-2118)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "ft-next-syndication-api",
   "description": "Next Syndication API",
-  "version": "0.41.6",
+  "version": "0.41.7",
   "private": true,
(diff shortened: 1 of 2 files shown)


Nazehs (approved):
looks okay to me

PR #423, opened 2023-01-03
Fix npm start on next-syndication-api

asugar13:
### Description
The `npm start` command was not working on `next-syndication-api`. Instead, we had to use `npm run start-local` which also didn't work because it was not getting environment variables from Vault. Now, this has been fixed by adding the commands that were previously used in `npm run start-local` as toolkit tasks (that do get the environment variables from Vault)
### Ticket
[Jira](https://financialtimes.atlassian.net/jira/software/c/projects/ACC/boards/1461?modal=detail&selectedIssue=ACC-2145&assignee=61486642cee20b0069c3b3c1)


.circleci/config.yml
```diff
@@ -11,7 +11,6 @@ jobs:
       - tool-kit/persist-workspace:
           path: .
```
asugar13, line 12: changes in this file have been autogenerated.

Nazehs (approved):
nice. looks good to me. weldone Asier for this piece of work

PR #438, opened 2023-03-14
ACC-2264: prevent error about possible null errors in the assets

danieleruiz:
### Description
Due to an error with salesforce data, there may be cases in which we receive the assets of a contract as null, causing the code to break and not allow users to enter.
[Slack](https://financialtimes.slack.com/archives/C3LRB6JCE/p1678197718180599)

### Ticket
[ACC-2264](https://financialtimes.atlassian.net/browse/ACC-2264)
0.41.10

[ACC-2264]: https://financialtimes.atlassian.net/browse/ACC-2264?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "ft-next-syndication-api",
-  "version": "0.41.9",
+  "version": "0.41.10",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "ft-next-syndication-api",
-      "version": "0.41.9",
+      "version": "0.41.10",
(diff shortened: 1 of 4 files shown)


AniaMakes (approved):
This should do the trick to fix our immediate issue of
```
TypeError: Cannot read properties of null (reading 'forEach')
    at /app/server/lib/get-contract-by-id.js:53:16
```
thank you for picking this up!

PR #439, opened 2023-03-21
add licence_id to logs

danieleruiz:
### Description
Add licence_id to log with null assets contract

0.41.11


binaryberry (approved):
Looks good to me but do you also need to open a PR in syndication-dl ? The PR template says: "Link to the next-syndication-dl PR which bumps the next-syndication-api version". Or because we're adding a log / it's a patch you think it wouldn't cause any issues?

PR #447, opened 2023-05-09
ACC-2363 added force refresh functionality

AniaMakes:
### Description
The database Salesforce data can be up to 24 hours out of date. This is because we only refresh the database data every 24 hours. But sometimes the client purchased an add-on and want to use it asap. 

This PR adds an endpoint that can be called to run forced update of the database. 
It also ads capacity for a custom key (so that Customer Support has a password that's different to all the internal Syndication password). 

You will notice there is no test file for this new controller. A) the original controller, which I pretty much copy-pasted, didn't have a test. B) Given the amount of things that happen - calls to Salesforce and calls to the database - I am not really sure where to start writing tests - I could try, but it would take me several days - at least - to have a test that works. 

### Ticket
https://financialtimes.atlassian.net/browse/ACC-2363

This doesn't impact download functionality, this is adding a stand-alone api endpoint.


danieleruiz (approved):
looks good to me!

PR #448, opened 2023-05-10
Log warning when contract has no assets

alexmuller:
In production we see an unexpected error because of the call to `contract.items.reduce()`, where `contract.items` can be null.

This causes the error:

```
Cannot read properties of null (reading 'reduce')
```

This commit logs a warning that the contract is malformed along with a bit of metadata, and returns the undecorated contract.

0.41.13

https://github.com/Financial-Times/next-syndication-dl/pull/139

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "ft-next-syndication-api",
   "description": "Next Syndication API",
-  "version": "0.41.12",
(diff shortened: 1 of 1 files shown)


AniaMakes (approved):
Thank you, this should help if this happens again.

PR #449, opened 2023-05-11
ACC-2363 remove force refresh endpoint and controller

AniaMakes:
### Description
Removes most of #447 - when testing that, I realised that a call to `https://www.ft.com/syndication/contracts/contractID` does refresh the database. 

We are keeping the secondary key functionality in the `api-key` middleware. 

### Ticket
https://financialtimes.atlassian.net/browse/ACC-2363


binaryberry (approved):
Look good to me, and I see the EXTERNAL_SYNDICATION_API_KEY had been added to Vault too.

PR #450, opened 2023-05-26
ACC-2379 added env check functionality

AniaMakes:
### Description
There is an issue with aws-sdk and slashes in secret keys 🤷🏻‍♂️ 

We had this incident https://financialtimes.slack.com/archives/C02SM4C7BDZ - and there was at least one other that got resolved much quicker. It relates to https://github.com/aws/aws-sdk-js/issues/86#issuecomment-476371303. It's not entirely clear why this is an issue, but it is, and I'd rather other people didn't run into this problem.

I made this PR https://github.com/Financial-Times/cp-vault-interface/pull/122 but the Reliability team wasn't too keen on it (and gave their reasons why) - and suggested that we should crash the app instead. We discussed this in the Accounts dev huddle and the consensus was that people are happy to crash the app, but it would be better if we made it a separate CI step so that it's clear what the issue is. 

This PR adds a check function and wires it into CI. The CI is no longer generated by ToolKit as a result - I had a long discussion via DMs with @ivomurrell who advised that sometimes ToolKit plugins are indeed a large overhead - as is the case here, and if we remove the top comment in the ci yaml file, that would allow us to _just make the changes we need_

### Ticket
https://financialtimes.atlassian.net/browse/ACC-2379

0.41.14

There will be a separate PR because it needs the same changes to the circle CI, so it's a bit bigger than just changing the version.


danieleruiz (approved):
thanks for that!

alexmuller (approved):
This looks right to me! 👍

One thing to bear in mind is that this works because the secrets are shared between the `continuous-integration` and `production` environments - which isn't ideally what we want, because it means that a bug in development or CI can take down production.

So at some point in the future we might want to separate the keys between environments and this step will no longer be sufficient. But for now it's good!

PR #461, opened 2023-08-21
ACC-2510: add current period start and endate for multiyears contracts

danieleruiz:
### Description
We need to add new current period date columns in contracts table using the salesforce new payload fields (orders).

This way we can modify after the query that counts the “current_downloads” in contract_asset_register_data and contract_asset_data tables in syndication database.

we need to check that this “current_period_start_date” and “current_period_end_date” are updated when the current period change.

We have a 24h cronjob that calls the salesforce endpoint and we should check if the current period has changed checking the period status, we should always update this columns with the dates of the current period to keep it updated. 
### Ticket
[ACC-2510](https://financialtimes.atlassian.net/browse/ACC-2510)
1.3.0

pending

[ACC-2510]: https://financialtimes.atlassian.net/browse/ACC-2510?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/db/pg/column_mappings/contracts.js
+++ b/db/pg/column_mappings/contracts.js
@@ -11,11 +11,11 @@ module.exports = exports = {
 			cite: 'endDate',
 			id: 'end_date'
 		}, {
-			cite: 'currentPeriodStartDate',
-			id: 'current_period_start_date'
+			cite: 'currentStartDate',
+			id: 'current_start_date'
 		}, {
-			cite: 'currentPeriodEndDate',
-			id: 'current_period_end_date'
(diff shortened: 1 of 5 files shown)


emortong (approved):
Yay!! Great, thank you. We just need to wait for Salesforce to make the changes in prod to be able to test with multiyear contracts.

And we need to confirm if single-year contract orders have an Active status and if so if the enddate and startdate coincide with the contract's enddate and startdate

PR #469, opened 2023-10-03
ACC-2702 Deprecate FTSession in favour of FTSession_s

emortong:
### Description
Customer products as a whole has been moving away from using `FTSession` in favour of `FTSession_s`. Previously we were checking for `FTSession_s` existence but we were decoding `FTSession` to move forward in user authentication. Dev tools to masquerade users on the FT like `dotcom-debugger` only work with `FTSession_s`. We tend to masquerade while debugging syndication so deprecating `FTSession` will help us take better of advantage of development tools as well as it will align us with the direction that CP is taking.

### Ticket
[ticket ACC-2702](https://financialtimes.atlassian.net/browse/ACC-2702)

1.4.0

https://github.com/Financial-Times/next-syndication-dl/pull/151


jamesr101 (approved):
Thanks a lot for doing this

PR #472, opened 2023-10-04
Add a fallback option for authRes.location to avoid error

emortong:
### Description
`.split()` method was erroring when `authResLocation` is null

1.4.2

https://github.com/Financial-Times/next-syndication-dl/pull/152


jamesr101 (approved):
Nice one

PR #488, opened 2024-02-15
bugfix: make variables more explicit and reuse them less

emortong:
### Description
bug in: server/lib/get-contract-by-id.js line 107
```
contract = reformatSalesforceContract(contract);
```
Since the contract returned by `getSalesforceContractByID(contractId);` assigned to the variable ` contract` was being reassigned by the contract returned from `reformatSalesforceContract(contract);` the `orders` object that came from SalesForce but is not needed in our database was being lost. Therefore when we did:
```
const activeOrder = contract.orders.find(order => order.status === 'Activated' && new Date(order.startDate) <= currentTimeInMilliseconds && new Date(order.endDate) >= currentTimeInMilliseconds);
```

No activeOrder was being found.

This PR makes variable names more explicit and reuses them less to avoid issues from mutations.

### Ticket
https://financialtimes.atlassian.net/browse/ACC-2934

2.0.6

### DIFF
--- a/server/lib/get-contract-by-id.js
+++ b/server/lib/get-contract-by-id.js
@@ -102,29 +102,30 @@ module.exports = exports = async (contractId, locals = {}) => {
 			return decorateContract(contract_data, locals.hasGraphicSyndication);
 		}
 	}
-	let contract = await getSalesforceContractByID(contractId);
-	if (contract.success === true) {
-		contract = reformatSalesforceContract(contract);
-		contract.last_updated = new Date();
-		if (contract.orders) {
+
+	let sfContract = await getSalesforceContractByID(contractId);
+	if (sfContract.success === true) {
+		let reformattedContract = reformatSalesforceContract(sfContract);
+		reformattedContract.last_updated = new Date();
+		if (sfContract.orders) {
 			const currentTimeInMilliseconds = new Date();
-			const activeOrder = contract.orders.find(order => order.status === 'Activated' && new Date(order.startDate) <= currentTimeInMilliseconds && new Date(order.endDate) >= currentTimeInMilliseconds);
-			contract.current_start_date = new Date(activeOrder.startDate);
-			contract.current_end_date = new Date(activeOrder.endDate);
+			const activeOrder = sfContract.orders.find(order => order.status === 'Activated' && new Date(order.startDate) <= currentTimeInMilliseconds && new Date(order.endDate) >= currentTimeInMilliseconds);
+			reformattedContract.current_start_date = new Date(activeOrder.startDate);
+			reformattedContract.current_end_date = new Date(activeOrder.endDate);
 		}
 
-		contract = pgMapColumns(contract, contractsColumnMappings);
+		const mappedContract = pgMapColumns(reformattedContract, contractsColumnMappings);
 
 		if (locals.MASQUERADING !== true) {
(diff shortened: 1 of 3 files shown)


server/lib/get-contract-by-id.js
```diff
@@ -102,29 +102,30 @@ module.exports = exports = async (contractId, locals = {}) => {
...
+	if (sfContract.success === true) {
+		let reformattedContract = reformatSalesforceContract(sfContract);
+		reformattedContract.last_updated = new Date();
+		if (sfContract.orders) {
 			const currentTimeInMilliseconds = new Date();
-			const activeOrder = contract.orders.find(order => order.status === 'Activated' && new Date(order.startDate) <= currentTimeInMilliseconds && new Date(order.endDate) >= currentTimeInMilliseconds);
-			contract.current_start_date = new Date(activeOrder.startDate);
-			contract.current_end_date = new Date(activeOrder.endDate);
+			const activeOrder = sfContract.orders.find(order => order.status === 'Activated' && new Date(order.startDate) <= currentTimeInMilliseconds && new Date(order.endDate) >= currentTimeInMilliseconds);
+			reformattedContract.current_start_date = new Date(activeOrder.startDate);
```
asugar13, line 113: **Question:** Did you mean to set the `current_start_date` of `sfContract` here instead like the previous code was doing?

server/lib/get-contract-by-id.js
```diff
@@ -102,29 +102,30 @@ module.exports = exports = async (contractId, locals = {}) => {
...
 			// purpose of this unclear but leaving it in, this should come back from get_contract_data if the licence ID already exists
 			// I think it is to fill in the licence ID and update the database from the user's licence ID which is retrieved earlier in the middleware, so that licence IDs are populated, however it does it in a way that breaks things if you are looking at another licence's data but not masquerading as that user (which only customer support can do)
 			if (locals && locals.licence) {
-				contract.licence_id = locals.licence.id;
+				mappedContract.licence_id = locals.licence.id;
 			}
 
 			// If you get the error, Cannot set property '#<anonymous>' of undefined, try refreshing
-			[contract_data] = await db.syndication.upsert_contract([contract]);
+			await db.syndication.upsert_contract([mappedContract]);
```
asugar13, line 128: **Question:** Do you not need to set [contract_data] here for the code below to work correctly?

server/lib/get-contract-by-id.js
emortong: no.. what I intended is for `sfContract` to remain as the salesforce contract without it being mutated. `reformattedContract` is the data we are preparing for the database, and then `mappedContract` is for the query to insert to the database.

server/lib/get-contract-by-id.js
emortong: I deleted it because I could not find how it was being used and it was making the variable states more difficult to keep track of, in the same logic for `server/controllers/get-contract-by-id.js` (line 51) we were not capturing the returned data either

emortong: line 136 of this same file then reassigns contract_data again

asugar13: Ok. I can see it's being re-assigned on line 136 anyways.

asugar13 (approved):

PR #492, opened 2024-02-29
BugFix: Add optional chaining to activeOrders

emortong:
### Description
Getting the activeOrder's start and end date is something that is specific to multiyear contracts. There will sometimes not be an active order so we should expect for these values to be undefined. Added optional chaining to avoid errors.

eslint's ecmaVersion was increased so that the linter supports optional chaining

v2.0.8


jamesr101 (approved):
👍

PR #496, opened 2024-09-05
chore: Remove Graphite healthchecks

GlynnPhillips:
### Description

Remove the custom Graphite Healtchecks

### Justification

The FT is phasing out its use of Graphite and by the end of Q3 we need to stop reading data from Graphite.

I have decided to delete these healthchecks instead of replacing them because I believe they are providing very little value.

The 401 healthcheck was constantly erroring because its far too easy to trigger by situations like users sessions ending. It was also constantly flapping and no one was actioning it, showing its low value.

The healthchecks for article/archive download spikes have never alerted so migrating and maintaining these in their current form feels like a very low reward. Especially as we are very unfamiliar with this product so theres a chance these are providing false confidence.

### Ticket
https://financialtimes.atlassian.net/browse/ACC-3298
https://financialtimes.atlassian.net/browse/ACC-3299
https://financialtimes.atlassian.net/browse/ACC-3300
https://financialtimes.atlassian.net/browse/ACC-3301

2.0.9


asugar13 (approved):
Approving although breaks my heart to see my first graphite health check go away.

PR #500, opened 2024-10-24
ACC-3368 Remove Snyk References

emortong:
Usage of Snyk at the FT is deprecated.We will not be renewing our licence after 19th December 2024, therefore we are removing all references to it.


Nazehs (approved):
LGTM!

PR #506, opened 2025-01-08
chore: update dep to support node v22

ytcleon:
### Description

This PR is to update the dependencies so they support node 22.x

### Ticket

[LIF-30](https://financialtimes.atlassian.net/browse/LIF-30)

N/A

[LIF-30]: https://financialtimes.atlassian.net/browse/LIF-30?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,9 +4,9 @@
   "version": "2.0.9",
   "private": true,
   "dependencies": {
-    "@dotcom-reliability-kit/crash-handler": "^2.1.1",
-    "@dotcom-reliability-kit/middleware-log-errors": "^1.2.5",
-    "@financial-times/n-es-client": "4.1.0",
+    "@dotcom-reliability-kit/crash-handler": "^4.1.11",
+    "@dotcom-reliability-kit/middleware-log-errors": "^4.2.5",
+    "@financial-times/n-es-client": "6.0.1",
     "@financial-times/n-express": "^31.9.4",
     "@financial-times/n-mask-logger": "7.2.0",
     "ajv": "^6.0.0",
--- a/package-lock.json
+++ b/package-lock.json
@@ -8,9 +8,9 @@
       "name": "ft-next-syndication-api",
       "version": "2.0.9",
       "dependencies": {
-        "@dotcom-reliability-kit/crash-handler": "^2.1.1",
-        "@dotcom-reliability-kit/middleware-log-errors": "^1.2.5",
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "ft-next-syndication-api",
   "description": "Next Syndication API",
-  "version": "2.0.9",
+  "version": "3.0.0",
   "private": true,
   "dependencies": {
     "@dotcom-reliability-kit/crash-handler": "^2.1.1",
```
asugar13, line 7: **Question:** Don't we want to update the consuming packages first?

emortong: +1 and dependencies, I think reliability kit has higher versions supporting nodev22

asugar13: I still haven't got my head around this @emortong -> if we update third party packages to their latest version and they are using node v22 features then this app will crash.

asugar13: I'm not sure if reliability-kit uses nodeV22 features yet or if it just supports them for now.

ytcleon: Good point. I am trying to follow [the procedure to deploy the download server](https://github.com/Financial-Times/next-syndication-dl#deploying-the-download-server). I think I need to do this in a separate PR right?

emortong: @asugar13 
> I still haven't got my head around this though -> if we update third party packages to their latest version and they are using node v22 features then I think this app will crash.

Why would it crash if this app is also running v22 and supports those features?

>I'm not sure if reliability-kit uses nodeV22 features yet or if it just supports them for now.

https://github.com/search?q=repo%3AFinancial-Times%2Fdotcom-reliability-kit+22.&type=code

ytcleon: I think the minimal approach would be updating the FT related package to their latest so all of them support node 22. For the 3rd party packages, quite a few of them received no support and the others can still support node 22, so I will keep them unchanged.

ytcleon: I will put the version bump of the API itself in another PR. The version bump is needed to update `next-syndication-dl`

asugar13 (approved):

package.json
asugar13: I've approved! I think there's "some" risk to update packages to v22 on a host app running an older version but since they're FT-repos the risk can be safely neglected as we know they're likely not using v22 features yet.

PR #528, opened 2025-01-15
Chore: Update README

ytcleon:
### Description

Currently some of the instructions in README use `.env` instead of Doppler and is not working. In addition it lacks some crucial details to connect to the production DB. This PR updates those instructions so others can properly set up a working local environment for the syndication API.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,7 +87,12 @@ You MIGHT need:
 
 If you need to use the database locally, set up the database by following the instructions in [next-syndication-db-schema](https://github.com/Financial-Times/next-syndication-db-schema).
 
-If you are using postgres in Docker, you will need to edit your `.env` file to set `DATABASE_HOST` to `192.168.99.100`
+If you are using postgres in Docker, you will need to edit your `custom-environment-variables.yaml` file to change `DATABASE_HOST` to `DATABASE_HOST_DOCKER` (variable already defined in Doppler).
+
+```yaml
+  DB:
+    host: "DATABASE_HOST_DOCKER"  # DATABASE_HOST
+```
 
 ### Running the API
 
--- a/README.md
+++ b/README.md
@@ -121,9 +121,26 @@ Once you have set up the projects you want to work on, and want to run all proje
 
 ### Running against prod DB (optional)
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -78,7 +87,12 @@ You MIGHT need:
...
 
-If you are using postgres in Docker, you will need to edit your `.env` file to set `DATABASE_HOST` to `192.168.99.100`
+If you are using postgres in Docker, you will need to edit your `custom-environment-variables.yaml` file to change `DATABASE_HOST` to `DATABASE_HOST_DOCKER` (variable already defined in Doppler).
+
+```yaml
+  DB:
+    host: "DATABASE_HOST_DOCKER"  # DATABASE_HOST
+```
 
 ### Running the API
```
asugar13, line 97: **Nitpick:** I think we should stop mentioning the .env file on line 87 as we noticed that this needs to be on doppler instead.

README.md
```diff
@@ -78,7 +87,12 @@ You MIGHT need:
 
 If you need to use the database locally, set up the database by following the instructions in [next-syndication-db-schema](https://github.com/Financial-Times/next-syndication-db-schema).
 
-If you are using postgres in Docker, you will need to edit your `.env` file to set `DATABASE_HOST` to `192.168.99.100`
+If you are using postgres in Docker, you will need to edit your `custom-environment-variables.yaml` file to change `DATABASE_HOST` to `DATABASE_HOST_DOCKER` (variable already defined in Doppler).
```
asugar13, line 90: **Nitpick:** Thanks for adding these docs. I wonder if anyone ever ran postgres in Docker to run syndication 😅

README.md
ytcleon: I thought we still need to edit the `.env` for next-router. Just checked Doppler and all the `syn-` env var are available. Will update this part.

README.md
ytcleon: I doubt that too. In the README of `next-syndication-db-schema` it mentions setting up the Postgres DB in Docker (or tried to, since that section is empty 🤷), so I supposed someone *might* want to run the DB in Docker.

PR #529, opened 2025-01-15
chore: Toolkit migration to v4.x.x

Nazehs:
### Description

This PR is migrating the package to dotcom toolkit v4.x.x

### Ticket
[Jira Ticket ](https://financialtimes.atlassian.net/browse/LIF-100)

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,18 +33,18 @@
     "xmldom": "^0.6.0"
   },
   "devDependencies": {
-    "@dotcom-tool-kit/backend-heroku-app": "^3.0.8",
-    "@dotcom-tool-kit/eslint": "^3.1.5",
-    "@dotcom-tool-kit/husky-npm": "^4.1.0",
-    "@dotcom-tool-kit/mocha": "^3.1.5",
-    "@dotcom-tool-kit/npm": "^3.2.2",
+    "@dotcom-tool-kit/backend-heroku-app": "^4.1.4",
+    "@dotcom-tool-kit/eslint": "^4.2.4",
+    "@dotcom-tool-kit/husky-npm": "^5.1.4",
+    "@dotcom-tool-kit/mocha": "^4.2.4",
+    "@dotcom-tool-kit/npm": "^4.2.4",
     "@financial-times/eslint-config-next": "^6.0.0",
     "@financial-times/n-test": "^8.0.0",
     "chai": "^4.0.2",
     "chai-http": "^4.3.0",
     "check-engine": "^1.10.1",
     "coveralls": "^2.13.1",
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -56,7 +57,7 @@
     "node-pandoc": "^0.3.0",
     "nodemon": "^3.1.9",
     "nyc": "^11.0.2",
-    "pm2": "^2.6.1",
+    "pm2": "^5.4.3",
```
asugar13, line 60: **Suggestion:** Same suggestion as before. This is bumping 3 major versions in the same PR that does the tool-kit migration.

package.json
```diff
@@ -104,7 +105,7 @@
     ]
   },
   "engines": {
-    "node": "18.x"
+    "node": "22.x"
```
asugar13, line 108: **Question:** Wasn't this on node 22 already? I'd suggest waiting until the node22 migration is done before merging the toolkit migration.

package.json
```diff
@@ -4,9 +4,9 @@
   "version": "2.0.9",
   "private": true,
   "dependencies": {
-    "@dotcom-reliability-kit/crash-handler": "^2.1.1",
-    "@dotcom-reliability-kit/middleware-log-errors": "^1.2.5",
-    "@financial-times/n-es-client": "4.1.0",
+    "@dotcom-reliability-kit/crash-handler": "^4.1.12",
+    "@dotcom-reliability-kit/middleware-log-errors": "^4.2.6",
+    "@financial-times/n-es-client": "6.0.1",
```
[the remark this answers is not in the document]
Nazehs: This is me upgrading internal dependencies on the app. it has nothing to do with toolkit working.

package.json
Nazehs: I bump up the pm 2 version because is the existing version was not working after the changes on node version and toolkit.

package.json
Nazehs: The initial changes for node migration were reverted with the assumption that it was the cause of the recent incident so i sneak this back since the only place that will be left after toolkit migration will the `package.json`

ytcleon: **Question**: Is the node 22 migration necessary for the tool kit migration?

package.json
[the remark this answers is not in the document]
ytcleon: **Suggestion**: If the upgrade on dependencies is not related to the toolkit upgrade I think it would be better to leave it in another separate PR (dependency upgrade for node 22 migration would be a good candidate since it needs to upgrade those dependencies as well)

ytcleon (approved):

PR #539, opened 2025-04-02
Migrate to Reliability Kit logger

rowanmanning:
### Description

This migrates the system to use Reliability Kit logger instead of the deprecated n-mask-logger. There are a few things in the custom `Logger` class in this system that are handled by Reliability Kit logger (e.g. baseLogData), however, I wanted to keep this simple and just extract the old dependency.

At some point in future you may want to refactor/reassess the need for such a comprehensive helper.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,10 +9,10 @@
       "version": "3.0.1",
       "dependencies": {
         "@dotcom-reliability-kit/crash-handler": "^5.0.0",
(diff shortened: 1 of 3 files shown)


ytcleon (approved):

ytcleon:
It is weird that the `ES_AWS_ROTATED_SECRET_ACCESS_KEY` secret has "/" in it. (I suppose the AWS secret key should not contain such character). I will try to check when the key is rotated in AWS.

PR #549, opened 2025-07-22
LIF 612 - Add support for Flourish graphics

manasiSantFT:
### Description
This PR introduces a temporary solution to support Flourish graphics in articles by ensuring they are included in the `content.embeds` array, even when not embedded.

Ideally, Flourish graphics should be parsed and included in the `content.embeds` section. Currently, they are not, which affects downstream rendering. This change bridges that gap until a more robust solution is implemented.

#### Code changes

1. `server/lib/enrich/article.js`

- Parses and extracts Flourish graphics from the HTML body.
- Constructs a fallback image URL for the Flourish content.
- Creates a properly structured embed object and appends it to the content.embeds array.
- Adds corresponding unit tests for this logic.

2.  `server/lib/builders/document-builder.js`

- Updates `removeNonSyndicatableImages` to preserve Flourish content and prevent it from being removed inadvertently.
- Includes unit tests to verify the updated logic.

### Ticket
https://financialtimes.atlassian.net/browse/LIF-612

### DIFF
--- a/server/lib/enrich/article.js
+++ b/server/lib/enrich/article.js
@@ -34,7 +34,7 @@ function extractFourishEmbeds(contentHTMLBody) {
 			type: 'http://www.ft.com/ontology/content/Graphic'
 		});
 		}
-		return flourishEmbeds
+		return flourishEmbeds;
 		
 	} catch(error){
 		return null;
@@ -58,9 +58,11 @@ module.exports = exports = function article(content, contract, graphicSyndicatio
 
 	const flourishEmbeds = extractFourishEmbeds(content.bodyHTML);
 	if(flourishEmbeds) {
+		content.embeds = content.embeds ?? [];
 		content.embeds.push(...flourishEmbeds);
-		content.contentStats.graphics += flourishEmbeds.length; 
-		
+		if(content.contentStats){
+			content.contentStats.graphics += flourishEmbeds.length; 
+		}
 	}
 
 	content.wordCount = getWordCount(content);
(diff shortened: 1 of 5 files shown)


server/lib/builders/document-builder.js
```diff
@@ -88,18 +88,31 @@ module.exports = exports = class DocumentBuilder {
...
 		const embedsMap = arrayToMap(this.content.embeds);
-
 		Array.from(this.contentDocument.getElementsByTagName('img')).forEach(
 			(el) => {
-				const imageType = el.getAttribute('data-image-type');
-
+				let isFlourishElement = false;
+				// identify flourish element
+				const elementSrc = el.getAttribute('src');
+				if(elementSrc?.includes('public.flourish.studio/')) {
```
joejeet, line 96: Comment/suggestion: I'll consider a more functional approach instead of imperative style when making changes in the DOM. We could simplify this to something like:

```
// Determine if it is a Flourish Element
const isFlourishElement = elementSrc?.includes('public.flourish.studio/');
```

```
// Determine image type
const imageType = isFlourishElement
? 'graphic'
: el.getAttribute('data-image-type');
```
This reduces multiple declarations and less to maintain.

server/lib/enrich/article.js
```diff
@@ -13,6 +13,34 @@ const {
...
 
+function extractFourishEmbeds(contentHTMLBody) {
+	try {
+		// Regex to extract Flourish IDs
+		const flourishIdRegex = /data-flourish-id="(\d+)"/g;
+
+		const flourishEmbeds = [];
+		let match;
+
+		while ((match = flourishIdRegex.exec(contentHTMLBody)) !== null) {
```
joejeet, line 24: Comment: Regex.exec has a state inside it, explained [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/exec#description). Maybe we can use a simpler approach which doesn't involve any states. 

Example:
```
for (const match of contentHTMLBody.matchAll(regex)) {
  // match[1] will be captured ID I believe
  ids2.push(match[1]);
}
```

PR #551, opened 2025-09-24
Adding hasFlourishGraphics flag to the article

manasiSantFT:
### Description
This PR adds hasFlourishGraphics flag to the article containing value `true` or `false`. 

#### Next steps 
This flag will be used by n-syndication to decide whether to show info banner about flourish graphics or not.

### Ticket
https://financialtimes.atlassian.net/browse/LIF-771

### DIFF
--- a/test/server/lib/enrich/article.js
+++ b/test/server/lib/enrich/article.js
@@ -83,36 +83,40 @@ describe(MODULE_ID, function () {
 	});
 
 	describe('Handling flourish content', () => {
-		context('if article doesn\'t contain flourish graphics then hasFlourishGraphics is set to false', () => {
-			const FLOURISH_CONTENT_ID = '42ad255a-99f9-11e7-b83c-9588e51488a0';
-			const item = require(path.resolve(`${FIXTURES_DIRECTORY}/content/${FLOURISH_CONTENT_ID}.json`));
-			item.lang = 'en';
-			item.extension = DEFAULT_FORMAT;
+		context('article doesn\'t contain Flourish graphics', () => {
+			it('sets hasFlourishGraphics to false', () => {
+				const FLOURISH_CONTENT_ID = '42ad255a-99f9-11e7-b83c-9588e51488a0';
+				const item = require(path.resolve(`${FIXTURES_DIRECTORY}/content/${FLOURISH_CONTENT_ID}.json`));
+				item.lang = 'en';
+				item.extension = DEFAULT_FORMAT;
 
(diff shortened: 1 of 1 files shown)


server/lib/enrich/article.js
```diff
@@ -49,6 +49,10 @@ module.exports = exports = function article(content, contract) {
 		(graphicEmbeds &&
 			graphicEmbeds.some((item) => item.canBeSyndicated === 'yes'));
 
+	// check whether content has flourish graphics or not
+	const flourishIdRegex = /data-flourish-id="(\d+)"/;
```
andygout, line 53: **nitpick:** This regex is a constant (rather than a variable) where the convention is to use uppercase snake case.


```suggestion
	const FLOURISH_ID_REGEX = /data-flourish-id="(\d+)"/;
```

test/server/lib/enrich/article.js
```diff
@@ -81,4 +81,38 @@ describe(MODULE_ID, function () {
 			});
 		});
 	});
+
+	describe('Handling flourish content', () => {
+		it('if article doesn\'t contain flourish graphics then hasFlourishGraphics is set to false', () => {
```
andygout, line 86: **suggestion:** If the test description includes "if" then it's a good indication that the condition it expresses is broken out into a `context` block:

```suggestion
		context('article doesn\'t contain Flourish graphics', () => {
				it('sets hasFlourishGraphics to false', () => {
```

test/server/lib/enrich/article.js
```diff
@@ -81,4 +81,38 @@ describe(MODULE_ID, function () {
...
+				allowed: {
+					rich_articles: true
+				}
+			}
+			underTest(item, contract);
+
+			expect(item.hasFlourishGraphics).to.equal(false);
+		});
+
+		it('if article contains flourish graphics then hasFlourishGraphics is set to true', () => {
```
andygout, line 102: **suggestion:** If the test description includes "if" then it's a good indication that the condition it expresses is broken out into a `context` block:


```suggestion
		context('article contains Flourish graphics', () => {
				it('sets hasFlourishGraphics to true', () => {
```

manasiSantFT: Right makes sense. Updated.

test/server/lib/enrich/article.js
andygout: Commits for these changes have not yet appeared.

andygout (approved):

